Issue #39, opened 2013-04-23
How to have custom/complex Datatype for api parameter


rkgarg:
How to have my custom datatype Model & Model Schema visible in Swagger UI similar to: http://petstore.swagger.wordnik.com/#!/user/createUser_post_1 ?

With https://github.com/wordnik/swagger-core/wiki/Datatypes , seems like model API has to be defined somewhere. 

Is there any code example with this implementation?

PR #49, opened 2013-07-08
Support description option body_param

### LINKED ISSUE #42: No way to specify a 'body' parameter for a PUT request
Because of the following code in grape-swagger.rb, around line 126:

``` ruby
paramType = path.match(":#{param}") ? 'path' : (method == 'POST') ? 'body' : 'query'
```

you cannot have a nice swagger UI for a `PUT` method with a text box for body. Should the above be amended to be something like this:

``` ruby
paramType = path.match(":#{param}") ? 
  'path' : (method == 'POST' || method == 'PUT') ? 'body' : 'query'
```

or should we introduce a DSL way to specify param type of 'body', something along the lines of

``` ruby
params do
  optional :body, body: true, desc: 'This parameter is the body of the request'
end
```

I am willing to submit a pull request once an acceptable approach is determined..

dgutov:
This is a proposal to fix #42. See the included commit message for details.

With this change, we can at least have textarea input for complex types that couldn't be represented with form fields.

The included parameter names and descriptions are lost. We could try to put them in some formatted way into `notes`, but since there's no guarantee that it supports markdown, the result would be messy.

Ideally, this should be eventually solved in #39.

### DIFF
--- a/lib/grape-swagger.rb
+++ b/lib/grape-swagger.rb
@@ -94,7 +94,8 @@ module Grape
                     :nickname   => route.route_method + route.route_path.gsub(/[\/:\(\)\.]/,'-'),
                     :httpMethod => route.route_method,
                     :parameters => parse_header_params(route.route_headers) +
-                      parse_params(route.route_params, route.route_path, route.route_method)
+                      parse_params(route.route_params, route.route_path, route.route_method,
+                                   route.route_body_param)
                 }
                 operations.merge!({:errorResponses => http_codes}) unless http_codes.empty?
                 {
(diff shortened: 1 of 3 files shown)


u2:
What now?

vanburg:
Am I right, we can't use something like this still?

swagger ui 2016-04-13 05-52-23

So we limited with ``curl -X POST --header 'Content-Type: application/x-www-form-urlencoded'``?
BTW, I even tried this:

swagger ui 2016-04-13 05-55-58

But content type header overwritten at the end with urlencoded one:
``curl -X POST --header 'Content-Type: application/x-www-form-urlencoded' --header 'Accept: application/json' --header 'XAuthToken: 123' -d 'name=123&venue=123' 'http://localhost:3000/api/v1/events.json'
``

LeFnord:

dgutov:
Actually, I'm not sure this approach should be preferred anymore.

From trying out the `swagger-1.2` branch, it seems like it does try to determine whether it can represent the call's parameters via forms, and if not, adds a body parameter.

That doesn't always work, though. For instance, it adds a body parameter for `requires :words, type: Array`, but if I specify the array's elements as hashes:

```rb
      requires :words, type: Array do
        requires :letter1, type: String
        requires :letter2, type: String
      end
```

then the resulting interface shows an obviously broken form with two fields (`words[][letter1]` and `words[][letter2]`), and no body parameter input area.

Anyway, it seems like fixing the edge cases like this can be a better way forward.

PR #65, opened 2013-10-30
safer conversion of grape accepted types to swagger types

jdoconnor:
This allows for the grape accepted param type 'Boolean' to be converted properly to a swagger type.  The issue is that the full type of a Boolean includes the Virtus namespace.  This can probably be extracted for other Virtus types


sethherr:
I believe this can be closed,  fixed from #168.

jdoconnor:
agree.

PR #86, opened 2014-02-04
Resources don't need versions or {format}, only routes do.

jeromegn:
I don't think it makes sense to add a version or a `.{format}` to resource names. They don't represent routes, only the name of a combination of resources.

I believe this section is only to link to deeper section. I'm learning as I'm doing this. :)


LeFnord:

PR #113, opened 2014-07-04
make sure the type is a string

qqshfox:


LeFnord:

PR #191, opened 2015-01-04
Generate default API base path

### LINKED ISSUE #107: API basePath is coming back with port
I am trying to serve the docs from a relative path for an app using HTTPS.  The Swagger JSON returns something like 

```
{
  "apiVersion":"v1",
  "swaggerVersion":"1.2",
  "resourcePath":"",
  "apis":[{
    "path":"/auth/user.{format}",
    "operations":[{
      "produces":["application/xml","application/json", "application/vnd.api+json","text/plain"],
      "notes":"",
      "summary":"",
      "nickname":"GET-auth-user---format-",
      "httpMethod":"GET",
      "parameters":[]
     }
  ]}],
  "basePath":"https://accounts.kyck.dev:80/api"}
```

`https` and port 80 are in direct conflict.  I have tried various options with `add_swagger_documentation`, including `include_base_url: false` and `root_base_path: false`

Not sure what I am doing wrong, but would love some help.

dskecse:
This generates the default API `basePath`, so that even when url scheme and port are in direct conflict in `production` or `staging` environment (in fact those, that differ from `development` and `test`), it works as expected. Resolves #107. Sorry it took so long. @dblock?

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -12,7 +12,7 @@ Metrics/AbcSize:
 # Offense count: 1
 # Configuration parameters: CountComments.
 Metrics/ClassLength:
-  Max: 397
+  Max: 401
 
 # Offense count: 5
 Metrics/CyclomaticComplexity:
@@ -26,7 +26,7 @@ Metrics/LineLength:
 # Offense count: 13
 # Configuration parameters: CountComments.
 Metrics/MethodLength:
-  Max: 359
+  Max: 362
 
 # Offense count: 4
 Metrics/PerceivedComplexity:
--- a/lib/grape-swagger.rb
+++ b/lib/grape-swagger.rb
@@ -316,6 +316,16 @@ module Grape
               end
             end
 
+            def generate_default_base_path
+              lambda do |request|
+                if %w(development test).include?(ENV['RAILS_ENV'] || ENV['RACK_ENV'])
+                  request.base_url
+                else
+                  "#{request.scheme}://#{request.host}"
+                end
+              end
+            end
+
             def hide_documentation_path
               @@hide_documentation_path
             end
@@ -328,7 +338,7 @@ module Grape
               defaults = {
                 target_class: nil,
                 mount_path: '/swagger_doc',
-                base_path: nil,
+                base_path: generate_default_base_path,
                 api_version: '0.1',
                 markdown: nil,
                 hide_documentation_path: false,
--- a/spec/non_default_api_spec.rb
+++ b/spec/non_default_api_spec.rb
@@ -495,9 +495,22 @@ describe 'options: ' do
     #   JSON.parse(last_response.body)["basePath"].should == "https://example.org:80"
     # end
 
-    it 'uses https schema in endpoint doc' do
-      get '/swagger_doc/something.json', {}, 'rack.url_scheme' => 'https'
-      expect(JSON.parse(last_response.body)['basePath']).to eq 'https://example.org:80'
+    context 'when in development environment' do
+      before { ENV['RAILS_ENV'] = 'development' }
+
+      it 'uses https schema with base_url in endpoint doc' do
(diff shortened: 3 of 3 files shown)


dblock:
My biggest issue with this is that it has Rails all over. Not everybody uses Rails. 

Probably a better way to do this is to add a Railtie and set things up under Rails from it.

dskecse:
Isn't `ENV['RAILS_ENV'] || ENV['RACK_ENV']` check insufficient? And I use the methods from `Rack::Request` class in here. Could you be more specific in the second line, please?

dblock:
It's probably sufficient, but it tightly couples the implementation details of the gem to Rails and there's a cleaner mechanism to do this, called a Railtie. There're quite a few intros online, like [this one](http://wangjohn.github.io/railties/rails/gsoc/2013/07/10/introduction-to-railties.html). Basically it's a class that only loads when Rails loads and in which you can do whatever you want. In that class we want to (ideally) be able to do something like this:

``` ruby
Grape::API.swagger.configure do |config|
   config.base_path = ...
end
```

dskecse:
Well, I agree it is tightly coupled to Rails. And thanks for the info on Railtie. I'll investigate it as soon as possible and prepare a more robust solution to this issue.

dblock:
I have a bigger question: why can't the base url always be "#{request.scheme}://#{request.host}" ?

dskecse:
We need to specify `request.base_url`, which is url scheme, a host and a port, for `development` and `test` environment, e.g. `http://localhost:3000`. You can see we're explicitly specifying the port number here. And the other one for `production` and the like, e.g. `https://example.com`, especially when url scheme and port are in direct conflict.

dskecse:
Won't it require additional config file? I mean, when writing something like this:

```ruby
Grape::API.swagger.configure do |config|
  config.base_path = app_config['grape-swagger']['base_path']
end
```
If not, how should we differentiate between Rails configuration and simple Rack configuration? And whether it introduce duplicated knowledge?

dblock:
For Rails it's pretty common to configure things from `config/initializers/gem-name.rb`, so all I am saying is that we should probably enable _that_.

dskecse:
One more question: should we invoke `Grape::API.swagger.configure` or `GrapeSwagger.configure` instead? The latter seems more natural and correct, doesn't it?

dblock:
If this is going to be global for all swaggers, then `GrapeSwagger.configure`, if it's going to be just for that API, then `Grape::API.swagger.configure`. Obviously it would be interesting to have both.

dm1try:
Guys for me it seems like a dirty workaround.

`grape-swagger` already use `request.base_url` to build `basePath`, see [this code](https://github.com/tim-vandecasteele/grape-swagger/blob/c9fb720bc9e191dd4faa47237425a33edb899552/lib/grape-swagger.rb#L335)

so the main problem described in https://github.com/tim-vandecasteele/grape-swagger/issues/107 why `Rack::Request#base_url` includes 80 port:
```
"basePath":"https://accounts.kyck.dev:80/api"
```
`rack` already [omit default ports for suitable schemes](https://github.com/rack/rack/blob/dc53a8c26dc55d21240233b3d83d36efdef6e924/lib/rack/request.rb#L319)

so port 80 seems invalid and pretty sure it's a configuration problem.
if you are using a proxy for your application server make sure it provides all needed HTTP headers because `rack` [uses them to get a port](https://github.com/rack/rack/blob/dc53a8c26dc55d21240233b3d83d36efdef6e924/lib/rack/request.rb#L92)

example of configuration problem - [nginx as proxy for rails app](http://stackoverflow.com/questions/10168155/nginx-proxy-https-to-http-on-non-standard-port)

dskecse:
I mostly agree with @dm1try. What can you say, @dblock?

dblock:
Sure, but I don't know what to about this PR?

dskecse:
I mean, should we close the PR?

dblock:
Yes, we all agree, I wasn't sure whether you were saying something else. Closing.

PR #196, opened 2015-01-17
If :type is omitted, see if it's available in :using

### LINKED ISSUE #195: Should "type" be picked up from the "using" option?
Currently you have to specify both `using` and `type`, e.g. `expose :widget, using: Entities::Widget, documentation: {type: 'Widget'}`. Seems redundant. Is there any reason why `type` shouldn't be inferred from `using` when it's missing? I can't think of a reason but thought I'd ask first. If there are no objections I'll submit a PR.

jhollinger:
Resolves #195


dblock:
Great, thanks.

PR #200, opened 2015-01-26
Treat Symbol as string in params parsing

ypresto:
`type: Symbol` is sometimes useful when accepting enum params.

Currently this library treats it as custom data type sent as body, but it must be same as string in layer of http.

This PR make Symbol treated as string in swagger doc (with only 2 lines!).

Thanks for great product :)


ypresto:
```
Class definition is too long. [399/397]
```

Oh, no... How do I fix it?

dblock:
For Rubocop, just run `rubocop --auto-gen-config`.

dblock:
This also needs a CHANGELOG entry please. 

You should write a higher level integration test, too (something about the API and the generated swagger doc).

ypresto:
Added test for generated doc, and CHANGELOG :)

dblock:
Looks great!

PR #206, opened 2015-01-30
Fixed the bug where 'is_array' in the return entity is ignored.

igormoochnick:
When the entity is marked with 'is_array', swagger will show it as array of entities. Example:

      desc 'Get all teams', {
                            is_array: true,
                            entity: API::Entities::TeamRecord
      }
      get '/' do
        teams = TeamsCatalog.teams.limit(100).to_a
        present teams, include_children: true
      end

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -12,11 +12,11 @@ Metrics/AbcSize:
 # Offense count: 1
 # Configuration parameters: CountComments.
 Metrics/ClassLength:
-  Max: 397
+  Max: 401
 
 # Offense count: 5
 Metrics/CyclomaticComplexity:
-  Max: 93
+  Max: 94
 
 # Offense count: 195
 # Configuration parameters: AllowURI, URISchemes.
@@ -26,11 +26,11 @@ Metrics/LineLength:
 # Offense count: 13
 # Configuration parameters: CountComments.
 Metrics/MethodLength:
-  Max: 359
+  Max: 362
 
 # Offense count: 4
 Metrics/PerceivedComplexity:
-  Max: 96
+  Max: 97
 
 # Offense count: 7
 Style/ClassVars:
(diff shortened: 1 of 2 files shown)


dblock:

dblock:
Bump @igormoochnick

calfzhou:
this doesn't work with grape new dsl, is there any way to use for new dsl?
```ruby
desc 'blah blah' do
  success API::Entities::TeamRecord
end
get '/' do
  records = ...
  present records, using API::Entities::TeamRecord
end
```

dblock:
@calfzhou If it doesn't work on HEAD, it's a legit feature request.

frodrigo:
I'm also interested in this pull request. What is missing for integration into master ? it's only the CHANGELOG ? ... I can do it if need !

dblock:
It needs tests and a changelog, please feel free to finish this @frodrigo.

dblock:
Fixed in https://github.com/tim-vandecasteele/grape-swagger/commit/768454bac676417ad1fce0d1da2bf82835fbc6c6.

birbird:
Does this feature work now?
I used code like
```
  desc 'Get all teams', { is_array: true, entity: API::Entities::TeamRecord }
```
but the Response Class does not show as a array

rhomeister:
Is this supported in the new DSL? See calfzhou's question: https://github.com/ruby-grape/grape-swagger/pull/206#issuecomment-78010784

Something like:

```
      desc 'Return a user' do
        array true
        success User
        failure [[401, 'Unauthorized', 'APIV2::Entities::ApiError']]
      end
```

LeFnord:
yeap, see: [#defining-an-endpoint-as-an-array](https://github.com/ruby-grape/grape-swagger#defining-an-endpoint-as-an-array)

rhomeister:
OK, thanks that works. However, it requires passing a Hash instead of using the DSL. If I use both, I get the following deprecation warning:

`[DEPRECATION] Passing a options hash and a block to `desc` is deprecated. Move all hash options to block.`

Example code:
```
      desc 'Return all accessible memberships', is_array: true do
        success Entities::Membership
      end
```

LeFnord:
it should also be working with a block

PR #207, opened 2015-01-31
Mark required elements inside group in mutually exclusive set as optional

mintuhouse:


lib/grape-swagger.rb
```diff
@@ -50,6 +50,22 @@ def combine_namespaces(app)
         end
       end
 
+      def get_non_nested_params(params)
+        params.each do |param, value|
+          is_nested_param = /^#{ Regexp.quote param }\[.+\]$/
+          params.each do |p, _|
+            if p.match(is_nested_param)
+              params[p][:required] = false unless value[:required].presence
```
dblock: This is altering the original `params`, I don't think we want that side effect.

spec/mutually_exclusive_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+      end
+      post '/groups' do
+        {}
+      end
+
+      add_swagger_documentation
+    end
+  end
+
+  it 'retrieves the documentation for group parameters' do
```
dblock: Maybe better describe the intent vs. what we were getting before?

lib/grape-swagger.rb
```diff
@@ -50,6 +50,22 @@ def combine_namespaces(app)
...
+            end
+          end
+        end
+
+        params.reject do |param, _|
+          is_nested_param = /^#{ Regexp.quote param }\[.+\]$/
+          params.keys.any? { |p| p.match is_nested_param }
+        end
```
dblock: You should be able to combine `each` and `reject`, there's a bunch of duplicate processing here, notably `is_nested_param`.

dblock:
Please update CHANGELOG, too. Thanks!

dblock:
I'm merging this. 

Do you see any reason not to use `deep_dup`, would make things a bit shorter?

```ruby
        dup_params = params.deep_dup # duplicate the params as we are going to modify them
        dup_params.reject do |param, value|
          is_nested_param = /^#{ Regexp.quote param }\[.+\]$/
          0 < params.count do |p, _|
            match = p.match(is_nested_param)
            dup_params[p][:required] = false if match && !value[:required]
            match
          end
        end
```

mintuhouse:
`deep_dup` is a rails method. I am not sure if everyone who uses grape-swagger use it along with rails. So avoided using it.

dblock:
Makes sense.

PR #208, opened 2015-01-31
Fix float value type

u2:
Fixed Float parameters, exposed as Swagger float types


dblock:
This needs tests and a CHANGELOG entry please.

u2:
ok.

dblock:
Good, merging.

PR #210, opened 2015-02-03
Fixed the range :values option, now exposed as enum parameters

u2:


dblock:
Thanks @u2, rock on!

u2

dblock:
The rbx error looks unrelated, something that Travis upgraded underneath us, merging.

PR #214, opened 2015-02-11
Allow callables for hidden setting

zbelzer:
I would like to expose a different set of API features per client, so having the documentation conditionally hidden is very useful to me.

Not sure if there's a more compact way to say the same thing.

Also, RubyCop is complaining because I added three lines. Is this configured limitation or am I just so lucky that we were at the exact boundary before?


dblock:
For rubocop run `rubocop --auto-gen-config`.

dblock:
Squash your commits please?

zbelzer:
Squashed. Rubocop updated and it looks like travis is happy.

dblock:
We should probably also document this in README, what do you think?

zbelzer:
Docs added.

dblock:
Merged, thank you.

swistaczek:
:+1:

synth:
This does not seem to allow conditional based on request headers or parameters, right? Seems like it would be a simple fix to pass them along. Essentially, I have endpoints that use a custom Oauth "trusted" scope and I only want to show endpoints when a token is used that has that scope. Or perhaps is there another way to show/hide endpoints based on matching oauth scope?

zbelzer:
I would imagine you just need to pass contextual information to `call` in each location. Not sure where whatever information you are using is kept, but you can figure out if it's available by putting in a breakpoint before `route.route_hidden.respond_to?(:call) ? route.route_hidden.call : route.route_hidden`

PR #217, opened 2015-02-18
Support for multiple entities

swistaczek:
Hey @dblock,
I have created this pull request to add support for multiple entities in grape-swagger. One of usage case is when I would like to specify some entity as required for input and I would like to return other entity in response for that query. I have described following case in specs.

Thanks in advantage for any clues if there is something to fix.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -7,7 +7,7 @@
 
 # Offense count: 8
 Metrics/AbcSize:
-  Max: 331
+  Max: 334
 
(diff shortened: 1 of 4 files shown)


dblock:
This is great, merged squashed via https://github.com/tim-vandecasteele/grape-swagger/commit/21f5f2a7320b8c6cf47d094b72a71c9de50835d7.

I think I would also like to be able to do both `entities: [ ]` and `entity: ...`, maybe you can make an additional PR for that?

PR #234, opened 2015-03-17
 Fixed the range :values options, now it does not fail with float range

### LINKED ISSUE #233: Params `values` option float range and enums
Setting param `values` option to float range makes the gem fail.

E.g. param:
`requires :field, type: Float, values: 1.0..10.0`

Exception & bt:
```
Rack app error: #<TypeError: can't iterate from Float>
/home/azhi/.rvm/gems/ruby-2.1.3/gems/activesupport-4.1.1/lib/active_support/core_ext/range/each.rb:7:in `each'
/home/azhi/.rvm/gems/ruby-2.1.3/gems/activesupport-4.1.1/lib/active_support/core_ext/range/each.rb:7:in `each_with_time_with_zone'
/home/azhi/.rvm/gems/ruby-2.1.3/gems/grape-swagger-0.10.1/lib/grape-swagger.rb:208:in `to_a'
/home/azhi/.rvm/gems/ruby-2.1.3/gems/grape-swagger-0.10.1/lib/grape-swagger.rb:208:in `block in parse_params'
/home/azhi/.rvm/gems/ruby-2.1.3/gems/grape-swagger-0.10.1/lib/grape-swagger.rb:179:in `each'
/home/azhi/.rvm/gems/ruby-2.1.3/gems/grape-swagger-0.10.1/lib/grape-swagger.rb:179:in `map'
/home/azhi/.rvm/gems/ruby-2.1.3/gems/grape-swagger-0.10.1/lib/grape-swagger.rb:179:in `parse_params'
/home/azhi/.rvm/gems/ruby-2.1.3/gems/grape-swagger-0.10.1/lib/grape-swagger.rb:571:in `block (3 levels) in setup'
/home/azhi/.rvm/gems/ruby-2.1.3/gems/grape-swagger-0.10.1/lib/grape-swagger.rb:555:in `map'
/home/azhi/.rvm/gems/ruby-2.1.3/gems/grape-swagger-0.10.1/lib/grape-swagger.rb:555:in `block (2 levels) in setup'
/home/azhi/.rvm/gems/ruby-2.1.3/gems/grape-swagger-0.10.1/lib/grape-swagger.rb:554:in `each'
/home/azhi/.rvm/gems/ruby-2.1.3/gems/grape-swagger-0.10.1/lib/grape-swagger.rb:554:in `block in setup'
/home/azhi/.rvm/gems/ruby-2.1.3/gems/grape-0.11.0/lib/grape/endpoint.rb:47:in `call'
/home/azhi/.rvm/gems/ruby-2.1.3/gems/grape-0.11.0/lib/grape/endpoint.rb:47:in `block in generate_api_method'
/home/azhi/.rvm/gems/ruby-2.1.3/gems/grape-0.11.0/lib/grape/endpoint.rb:247:in `call'
/home/azhi/.rvm/gems/ruby-2.1.3/gems/grape-0.11.0/lib/grape/endpoint.rb:247:in `run'
/home/azhi/.rvm/gems/ruby-2.1.3/gems/grape-0.11.0/lib/grape/endpoint.rb:195:in `block in call!'
...
```

This behaviour was introduced in https://github.com/tim-vandecasteele/grape-swagger/commit/2ac951f42c3c1471d0d1ec27ad2979101880ddf5#diff-3bdef0e908741c1b55e8e9660610e33dR101. 
My guess is parsing `values` range into enum should only be done if value in range responds to `:succ`, or even explicitly whitelist range value types to `Integer` and `String`.

azhi:
now range values is exposed as array in enum field only if it is Integer or
 String range
 float range or procs returning non-array values is exposed as string
 fixes #233


azhi:
Grape below 0.11.0 has same bug: it tried to iterate through range without checking if it can be done.
It was fixed with some values validating regression fix (see https://github.com/intridea/grape/commit/0b98900f09af8a55c9c093ca38a35977f948de1d).
Build for all grape versions below 0.11.0 will fail.

dblock:
The tests for these older versions of Grape should be skipped. I think you can just condition the tests on the version of Grape and I would be ok to merge.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 #### Fixes
 
 * [#232](https://github.com/tim-vandecasteele/grape-swagger/pull/232): Fixed missing raw array params - [@u2](https://github.com/u2).
+* [#234](https://github.com/tim-vandecasteele/grape-swagger/pull/234): Fixed the range :values options, now it does not fail with float range - [@azhi](https://github.com/azhi).
```
dblock: Maybe simpler: "Fixed range :values with float" or something like that?

lib/grape-swagger.rb
```diff
@@ -160,6 +160,13 @@ def parse_array_params(params)
         modified_params
       end
 
+      def parse_enum_values(values)
+        enum_values = values
+        enum_values = enum_values.to_a if enum_values && enum_values.is_a?(Range) && [Integer, String].any? { |klass| enum_values.first.is_a?(klass) }
+        enum_values = enum_values.call if enum_values && enum_values.is_a?(Proc)
```
dblock: These are mutually exclusive, right? And `enum_values` is unnecessary, so try:

```ruby
if values && values.is_a?(Range) ...
   values = values.to_a
elsif values.is_a?(Proc)
  ...
end
```

dblock:
As you pointed out in #233 you cannot be converting ranges into arrays - it would be eating a lot of memory and take a very long time. We need a better solution.

azhi:
Updated.

@dblock i think the proper solution would be not to convert ranges to enums at all. However, this feature was in the release and some people might be using it.

Also we could try some dirty hack like convert to an array only if Range#size is less then some  amount (say, 50). But that seems very odd to me.

dblock:
I am going to merge this. Could you open a new issue regarding the Range problem? Thanks.

PR #238, opened 2015-03-31
Fix nested entity names in parameters and as in models.

frodrigo:
Fix nested entity names in parameters and as in models.


lib/grape-swagger.rb
```diff
@@ -210,7 +210,7 @@ def parse_params(params, path, method)
                                 when 'Symbol'
                                   'string'
                                 else
-                                  @@documentation_class.parse_entity_name(raw_data_type)
+                                  @@documentation_class.parse_entity_name(value.is_a?(Hash) ? value[:type] || 'string' : 'string')
```
dblock: Maybe clearer:

```ruby
raw_data_type = value[:type] if value.is_a?(Hash)
raw_data_type ||= 'string'
```

dblock:
Thanks! Fix build, see my minor comment above and squash the commits, please.

lib/grape-swagger.rb
frodrigo: No, because the first one is with .to_s and not the last

frodrigo:
I stashed fix for my changes, but there is issues with the main code

dblock:
You should run `rubocop --auto-gen-config`, squash your commits and this should be good to merge.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 * [#215](https://github.com/tim-vandecasteele/grape-swagger/pull/223): Support swagger `defaultValue` without the need to set a Grape `default` - [@jv-dan](https://github.com/jv-dan).
 * [#225](https://github.com/tim-vandecasteele/grape-swagger/pull/225): Fixed param_type to have it read from parameter's documentation hash - [@zsxking](https://github.com/zsxking).
+* [#235](https://github.com/tim-vandecasteele/grape-swagger/pull/235): Fix nested entity names in parameters and as $ref in models - [@frodrigo](https://github.com/frodrigo).
```
dblock: Make it quoted `$ref` just for consistency.

lib/grape-swagger.rb
```diff
@@ -191,8 +191,9 @@ def parse_params(params, path, method)
               non_nested_parent_params.map do |param, value|
                 items = {}
 
-                raw_data_type = value.is_a?(Hash) ? (value[:type] || 'string').to_s : 'string'
-                data_type     = case raw_data_type
+                raw_data_type = value[:type] if value.is_a?(Hash)
```
dblock: Nitpicking, since you're doing a `to_s` below you could:

```ruby
raw_data_type = value[:type] if value.is_a?(Hash)
raw_data_type = raw_data_type ? raw_data_type.to_s : 'string'
```

spec/reference_entity.rb
```diff
@@ -0,0 +1,80 @@
+require 'spec_helper'
+
+describe 'referenceEntity' do
+  before :all do
```
dblock: No need to change this, but eventually I'd like to change this whole thing into something that does `let(:something) { Class.new(Grape::Entity) do ... }` this way we're not polluting global namespaces with custom names.

lib/grape-swagger.rb
frodrigo: No, there is no to_s in all case, thaht the point of my fix !

dblock:
Merged via https://github.com/tim-vandecasteele/grape-swagger/commit/eb3553de398ef92b01acb42742c69509519eb4d0. Thank you.

PR #240, opened 2015-04-03
swagger_doc.json -> swagger_doc

whatasunnyday:
The README states that the path mounted by default will be swagger_doc.json but it's swagger_doc according to: https://github.com/tim-vandecasteele/grape-swagger/blob/master/lib/grape-swagger.rb#L470.


dblock:
Correct.

PR #242, opened 2015-04-06
Update README.md

TangMonk:
Add swagger-ui version requirement

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,10 @@ Add to your Gemfile:
 
 Please see [UPGRADING](UPGRADING.md) when upgrading from a previous version.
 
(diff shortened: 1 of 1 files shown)


dblock:
Having looked at this more, there's no reason why a newer swagger-ui shouldn't work. They say they are backwards compatible with swagger-spec 1.2, so there is probably a bug somewhere? Care to debug it?

I added a note that grape-swagger generates 1.2 documentation in https://github.com/tim-vandecasteele/grape-swagger/commit/e77cbbd05b92105a2759c99f78905a2230edd8db, that is definitely useful, thanks for suggesting it.

PR #245, opened 2015-04-08
Set default format to be .json.

sunnyrjuneja:
Fixes: 
https://github.com/tim-vandecasteele/grape-swagger/issues/244
https://github.com/TinkerDev/grape-swagger-rails/issues/11

Perhaps more. Not sure if we need any specs for this.

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -1,6 +1,17 @@
 Upgrading Grape-swagger
 =======================
 
+### Upgrading to >= 0.10.2
+
+Previously, grape-swagger hid the format in the URL by default. This conflicted
+with grape-swagger-rails that required grape-swagger to specify the format (see
+below). Grape-swagger will now assume the `format :json` by default. To
+revert, `add_swagger_documentation(format: nil)`
+
+#### Grape-Swagger-Rails
+
+You will no longer need to specify `add_swagger_documentation(format: :json).`
+
 ### Upgrading to >= 0.9.0
 
 #### Grape-Swagger-Rails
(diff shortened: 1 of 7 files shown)


dblock:
I definitely think we should have specs, some behavior has changed, right?

dblock:
It also needs a CHANGELOG and an UPGRADING entry.

dmitry:
Since that change is it possible to access to the resources with the `.json` extension postfix?

dmitry:
@whatasunnyday can you please update changelog and upgrading files?

whatasunnyday:
@dmitry Sure.

dmitry:
@whatasunnyday :+1: 
@dblock I believe everything is ready to be merged now.

UPGRADING.md
```diff
@@ -1,6 +1,17 @@
 Upgrading Grape-swagger
 =======================
 
+### Upgrading to >= 0.10.2
+
+Previously, grape-swagger hide the format in the URL by default. This conflicted
```
dblock: `hide` => `hid`

UPGRADING.md
```diff
@@ -1,6 +1,17 @@
...
+with grape-swagger-rails that required grape-swagger to specify the format (see
+below). Now, grape-swagger will now assume the `format :json` by default. To
```
dblock: Gape-swagger will now assume ...

dblock:
I still think this needs a test! We're changing behavior, what prevents a future committer to revert it?

whatasunnyday:
I actually agree 100% but I'm quite confused the current state of the tests. The tests fail for grape 0.10.0 to 0.11.0 but pass on HEAD. Why is this?

I was thinking to change https://github.com/tim-vandecasteele/grape-swagger/blob/master/spec/default_api_spec.rb, specifically:

```ruby
    subject do
      get '/swagger_doc'
      JSON.parse(last_response.body)
    end
```

to

```ruby
    subject do
      get '/swagger_doc.doc'
      JSON.parse(last_response.body)
    end
```

but the tests pass even though I changed `format: :json`. I do not understand if the tests are testing anything at all!

whatasunnyday:
Also, the tests here should be folded into default_api_spec.rb.

https://github.com/tim-vandecasteele/grape-swagger/blob/master/spec/non_default_api_spec.rb#L583-L612

dblock:
Right, we are likely missing tests around this, this is a good time to add them.

With your change Grape 0.10.0, 0.11.0, etc. are broken by default. So we can't ship this, it looks like it's as simple as requesting with .json vs. without .json, but I am not 100% sure what to do about it. Either way we want this PR not to break any tests ;)

LeFnord:

Issue #250, opened 2015-04-22
Models not following swagger spec 1.2 when Grape entity contains root


aitortomas:
When using Grape entities, if `root`is defined, the spec returned may not follow the swagger spec 1.2 for the ID field.
As per https://github.com/swagger-api/swagger-spec/blob/master/versions/1.2.md#527-model-object the ID field "MUST be the name given to {Model Name}".
When using Grape entities if root is defined and different than model name, the spec generated does not follow the above rule.
As an example:

```
class AutosEntity < Grape::Entity
    root 'cars', 'car'
    ...
end
```

will generate:

```
...
"AutosEntity":{  
         "id":"car",
...
```

This is due to:
https://github.com/tim-vandecasteele/grape-swagger/blob/master/lib/grape-swagger.rb#L386

It causes that swagger-ui shows as schema:

```
Inline Model {
undefined
}
undefined {
}
```

dblock:
Would appreciate a PR with a spec and maybe even a fix?

LeFnord:
solved in the actual version

PR #251, opened 2015-04-23
Id equal to model name when root is used in Entity

aitortomas:
See issue #250

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 #### Fixes
 
+* [#251](https://github.com/tim-vandecasteele/grape-swagger/pull/251): Fixed model id equal to model name when root existing in entities - [@aitortomas](https://github.com/aitortomas).
 * [#232](https://github.com/tim-vandecasteele/grape-swagger/pull/232): Fixed missing raw array params - [@u2](https://github.com/u2).
 * [#234](https://github.com/tim-vandecasteele/grape-swagger/pull/234): Fixed range :values with float - [@azhi](https://github.com/azhi).
 * [#225](https://github.com/tim-vandecasteele/grape-swagger/pull/225): Fixed `param_type` to have it read from parameter's documentation hash - [@zsxking](https://github.com/zsxking).
--- a/lib/grape-swagger.rb
+++ b/lib/grape-swagger.rb
@@ -334,6 +334,8 @@ module Grape
             def parse_entity_name(model)
               if model.respond_to?(:entity_name)
                 model.entity_name
+              elsif model.instance_variable_get(:@root)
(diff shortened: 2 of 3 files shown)


aitortomas:
With my fix, the schema in Swagger-UI would show with the class name from the entity.
Should it show the defined "root" if existing?
I am open to discussion.

dblock:
So this was a misunderstanding of the spec?

Fix the build please (rubocop), update changelog.

aitortomas:
Yes, it looks like a misunderstanding of the specs, at least for the case in which root is defined in the entities.

dblock:
Merged via https://github.com/tim-vandecasteele/grape-swagger/commit/834ee0111e2631e8359f429c9184207579aa940c.

iangreenleaf:
This now breaks my entity references because `root` gets used in the `models` object but not in `apis[].operations[].type`.

Should `@root` be used at all by default? Seems like it's internal data and meant for defining the formatting, not for defining the model ID.

aitortomas:
I see what you mean, I can put the root logic inside the `parse_entity_name` method and it will work.
Despite that, I am open to discussion regarding if root should actually be used or not.

dblock:
@iangreenleaf and @aitortomas Can you please help resolve this? Lets make sure we don't break backward compatibility or have a way out documented in UPGRADING?

aitortomas:
I can add the root logic to the `parse_entity_name` method but then Rubocop complains:
lib/grape-swagger.rb:9:3: C: Class definition is too long. [511/509]
  class API
  ^^^^^
lib/grape-swagger.rb:173:7: C: Assignment Branch Condition size for create_documentation_class is too high. [360.7/360]
      def create_documentation_class
      ^^^
lib/grape-swagger.rb:173:7: C: Method has too many lines. [383/381]
      def create_documentation_class
      ^^^

aitortomas:
Check https://github.com/tim-vandecasteele/grape-swagger/pull/254

iangreenleaf:
Looks like #254 is a good fix if we're keeping the root logic. I'm still on the side of removing it, though. @dblock @aitortomas do either of you know why it was added in the first place? I found the commit that introduced it (2747bb2c) but I don't know if there's an associated issue and discussion.

It seems like the kind of behavior that might be convenient for some apps, but not everyone in general. The fact that I want my output wrapped as `{xyz: {…}}` doesn't necessarily mean I want `xyz` to be my canonical model name.

Removing this *would* break backwards compatibility with anyone depending on it, but I think we could work out a simple upgrade path. For example, people who wanted this behavior could redefine `entity_name` to something like this:

```ruby
def entity_name
  super || @root
end
```

dblock:
To answer your question, honestly I don't know the background for this.

PR #252, opened 2015-04-28
Allow docs to be generated in separate class from target

iangreenleaf:
I think this is the intention of the `target_class` option—to allow the docs for an API class to be generated in a different class. The existing implementation didn't support the full split, so I did a quick fix to change that. You can take a look at the spec I added to check that my understanding of the desired behavior is correct.


dblock:
Please update CHANGELOG, fix the build and squash the commits (`rubocop -a` will do the job). I'll take a closer look at the code.

iangreenleaf:
@dblock Done. Rubocop made a bunch of random code changes, and forced me to refactor for length as well. So, fair warning, this is now merge conflict bonanza.

lib/grape-swagger.rb
```diff
@@ -172,478 +175,480 @@ def parse_enum_values(values)
...
+        end
+      end
+    end
+  end
+end
 
-            def as_markdown(description)
-              description && @@markdown ? @@markdown.as_markdown(strip_heredoc(description)) : description
-            end
+module GrapeSwaggerDocMethods
```
dblock: Can we pull this one out into its own file?

dblock:
Looks good, see my comment above and please squash these commits, too.

iangreenleaf:
Ok, pulled the module out and squashed. I didn't squash the Rubocop refactoring, since those commits are totally unrelated to the actual change. That okay with you?

dblock:
This looks great, merging.

PR #254, opened 2015-04-29
Fix for Id equal to model name when root is used in Entity

aitortomas:
As per https://github.com/tim-vandecasteele/grape-swagger/pull/251
This adds the root logic to `parse_entity_name` method.


aitortomas:
@iangreenleaf can you check now?

iangreenleaf:
@aitortomas I'll test later today, but this should fix the issue, since it makes the naming consistent.

aitortomas:
Great @iangreenleaf.
I did it like this the first time and the change to comply with Rubocop.
My mistake.

iangreenleaf:
Actually, something that just occurred to me: this isn't fully backwards compatible either. The old behavior would override everything if `@root` was defined. The new behavior defers to `entity_name` if that's defined. I think the new behavior is *much better*, but it's still a breaking change.

aitortomas:
You are right.
I also agree with your comment in #251, I do not know why the root logic was added in the first place as you can always rely on `entity_name`.
Considering it a breaking change, maybe it is better to remove the root logic then, which seems "less breaking".

iangreenleaf:
> Considering it a breaking change, maybe it is better to remove the root logic then, which seems "less breaking".

I totally agree. If we're going to meddle no matter what, we might as well make the best change.

aitortomas:
finally I removed the root logic completely. as stated by @iangreenleaf you can always use `entity_name` to add as model name the root value.

dblock:
This definitely needs an UPGRADING paragraph.

Please rebase & squash the commits, too.

aitortomas:
I have been thinking about the UPGRADING paragraph and it might be confusing. I mean, I doubt nobody was using the `root` in entities without having `entity_name` defined as it wold have not work with swagger-ui. So, removing the root logic as we did, in fact, should not affect anyone which has everything running correctly. What do you think @iangreenleaf?

iangreenleaf:
You might be right. If anyone is currently using the `@root` logic, would they be gaining something useful from it that will go away with this change? Or would it only be buying them a mysterious breakage that this will fix?

dblock:
Sounds like you both agree, I am happy to merge this, it needs a squash + rebase.

dblock:
Bump re: rebase / squash.

LeFnord:

PR #256, opened 2015-05-12
more robust logic for cleaning out the format suffix when hide_format is specified

mzruya:
`hide_format: true` only works in a specific case
the suffix of the path could be one of three options (one of which depends on the actual format)

`grape/path.rb:39`
```ruby
if uses_specific_format?
  "(.#{settings[:format]})"
elsif !uses_path_versioning? || (has_namespace? || has_path?)
  '(.:format)'
else
  '(/.:format)'
end
```

Since we only have the already interpolated path string, we can figure out which option was the one using a regex.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 #### Fixes
 
+* [#256](https://github.com/tim-vandecasteele/grape-swagger/pull/256): More robust logic for cleaning out the format suffix when hide_format is specified - [@mzruya](https://github.com/mzruya).
 * [#252](https://github.com/tim-vandecasteele/grape-swagger/pull/252): Allow docs to mounted in separate class than target - [@iangreenleaf](https://github.com/iangreenleaf).
 * [#251](https://github.com/tim-vandecasteele/grape-swagger/pull/251): Fixed model id equal to model name when root existing in entities - [@aitortomas](https://github.com/aitortomas).
 * [#232](https://github.com/tim-vandecasteele/grape-swagger/pull/232): Fixed missing raw array params - [@u2](https://github.com/u2).
--- a/spec/non_default_api_spec.rb
+++ b/spec/non_default_api_spec.rb
@@ -504,22 +504,40 @@ describe 'options: ' do
           { bla: 'something' }
         end
       end
+    end
 
-      class SimpleApiWithHiddenPaths < Grape::API
-        mount ProtectedApi
-        add_swagger_documentation hide_format: true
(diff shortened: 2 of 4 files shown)


dblock:
Can you please take a look at the spec failures first? Thanks.

dblock:
Bump @mzruya

mzruya:
@dblock, sorry about that, my change was a bit too naive, fixed it.

dblock:
Thanks, this still needs a test and a CHANGELOG entry.

mzruya:
I just noticed that the gem is using an older version of grape, the snippet of code I put above is from the latest and has evolved a bit, do you mind if I update the Gemfile to the latest released version? 
that way I can test against the different cases.

mzruya:
@dblock, added specs and a CHANGELOG entry

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 
 #### Fixes
 
+* [#256](https://github.com/tim-vandecasteele/grape-swagger/pull/256): More robust logic for cleaning out the format suffix when hide_format is specified - [@mzruya](https://github.com/mzruya)
```
dblock: OCD: Missing a period at the end.

spec/non_default_api_spec.rb
```diff
@@ -497,7 +497,9 @@ def app
   end
 
   context ':hide_format' do
-    before :all do
+    let(:doc_options) { Hash.new }
+    before do
+      stub_const('DOC_OPTIONS', doc_options)
```
dblock: This is add, can you explain?

dblock:
Thanks. See my questions & comments, and squash the commits, please.

spec/non_default_api_spec.rb
mzruya: What I tried achieving here is providing different options for `add_swagger_documentation` in `let` blocks, while still having one setup code for all specs (see `doc_options`).

Because the method is invoked inside the class definition, the only way I could pass something to it is via a defining a constant in a scope where the let's can be accessed, then in turn the constant is available in the class definition.

It's a bit hacky, I would love for a way to achieve the same without resorting to trickery :-)

dblock: `shared_examples_for` can take an argument, and then instead of defining a class like `SimpleApiWithHiddenPaths`, define that dynamically: 

```ruby
Class.new(Grape::API) do
 ...
end
```

Does this help?

dblock:
Did you mean to close this without merging?

PR #259, opened 2015-06-03
Converting integer :values range to a minimum/maximum numeric Range

u2:
https://github.com/swagger-api/swagger-core/wiki/Datatypes

>Some model properties may be restricted to a fixed set of values. To support this use case, the property field can have an optional enum property, which supports a list of allowable values, or a minimum/maximum numeric Range. For example:
```
"Pet": {
  "id": "Pet",
  "description": "A pet is a person's best friend",
  "required": [
    "name",
    "id"
  ],
  "properties": {
    "status": {
      "type": "string",
      "description": "pet status in the store",
      "enum": [
        "available",
        "pending",
        "sold"
      ]
    },
    "happiness": {
      "type": "integer",
      "format": "int32",
      "description": "how happy the Pet appears to be, where 10 is 'extremely happy'",
      "minimum": 1,
      "maximum": 10
    },
  ...
```

So the `Ineteger` `Range` should be converted to `minimum/maximum numeric Range`, right?

### DIFF
--- a/spec/param_values_spec.rb
+++ b/spec/param_values_spec.rb
@@ -1,7 +1,7 @@
 require 'spec_helper'
 require 'grape_version'
 
-describe 'Convert values to enum' do
+describe 'Convert values to enum or Range' do
   def app
     Class.new(Grape::API) do
       format :json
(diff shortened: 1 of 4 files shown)


spec/param_values_spec.rb
```diff
@@ -64,17 +64,17 @@ def first_parameter_info(request)
   context 'Range values' do
     subject(:range_letter) { first_parameter_info('range_letter') }
 
-    it 'has letter range values as array in enum' do
+    it 'has letter range values' do
       expect(range_letter).to eq [
-        { 'paramType' => 'form', 'name' => 'letter', 'description' => nil, 'type' => 'string', 'required' => true, 'allowMultiple' => false, 'enum' => ('a'..'z').to_a }
```
dblock: Why would `enum` be incorrect here?

u2:
The `string` range can be converted to `string`, and the `Integer` range can be converted to `a minimum/maximum numeric Range`.As this, it would avoid the large range problem,but it's not perfect.

dblock:
I see. I think I want to merge this, CHANGELOG?

u2:
updated.

PR #265, opened 2015-06-24
Update to Ruby 2.2.2

tmilewski:


dblock:
Why do we even need this file? I think it should be removed and 2.2.2 added to https://github.com/tim-vandecasteele/grape-swagger/blob/master/.travis.yml, no?

mattolson:
Agreed, use of `.ruby-version` is really a local concern. Rather than changing it, It should be added to `.gitignore`.

dblock:
I would take a PR, please.

dblock:
Closing in favor of https://github.com/tim-vandecasteele/grape-swagger/pull/288.

PR #266, opened 2015-06-26
Respect primitve mapping on type and format attributs of 1.2 spec

frodrigo:
Respect primitve mapping on type and format attributs of 1.2 swagger spec: https://github.com/swagger-api/swagger-spec/blob/master/versions/1.2.md#431-primitives

### DIFF
--- a/lib/grape-swagger/doc_methods.rb
+++ b/lib/grape-swagger/doc_methods.rb
@@ -1,5 +1,15 @@
 module GrapeSwagger
   module DocMethods
+    PRIMITIVE_MAPPINGS = {
+      'integer' => %w(integer int32),
+      'long' => %w(integer int64),
+      'float' => %w(number float),
+      'double' => %w(number double),
+      'byte' => %w(string byte),
+      'date' => %w(string date),
+      'dateTime' => %w(string date-time)
+    }
+
     def name
       @@class_name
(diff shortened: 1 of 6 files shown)


lib/grape-swagger/doc_methods.rb
```diff
@@ -1,5 +1,15 @@
 module GrapeSwagger
   module DocMethods
+    @@primitive_mapping = {
```
dblock: This can be a CONSTANT.

dblock:
Thanks. This needs tests and a passing build, please.

frodrigo:
All known problem fixed.

lib/grape-swagger/doc_methods.rb
```diff
@@ -1,5 +1,15 @@
 module GrapeSwagger
   module DocMethods
+    PRIMITIVE_MAPPING = {
```
dblock: Maybe PRIMITIVE_MAPPINGS (plural)?

dblock:
It would be nice to extend tests to all primitive mappings, but I would merge without. Can you squash the commits, please?

frodrigo:
Squash and plural done.

dblock:
Merged, thanks.

Issue #237, opened 2015-03-30
Array declaration is wrong


Fryie:
A param declaration like ```requires :foo, type: Array[Integer]``` generates a ```type: '[Integer]'``` entry in the JSON output.

However, the correct format would be:
```
type: 'array',
items: { type: 'integer' }
```
as [documented by the specification](https://github.com/swagger-api/swagger-spec/blob/master/versions/1.2.md#433-data-type-fields) - this btw doesn't appear to have changed in Swagger 2.0.

See also [this issue](https://github.com/swagger-api/swagger-ui/issues/982).

It probably should also automatically append "[]" to the parameter name, although I'm not sure about that...

dblock:
Check whether this is the case on HEAD and write a test that reproduces this? Thanks.

Martin91:
I have meet the same issue. Can you resolve it? @Fryie

Fryie:
Haven't had time to dig into it deeper, sorry. If you want, you can try to write a failing test as @dblock asked; otherwise I'll try to get to it as soon as I can manage.

frodrigo:
Can be closed #237

dblock:
Closed.

PR #268, opened 2015-06-30
Patch to handle Array Declaration #237

frodrigo:
Patch to handle Array Declaration #237

### DIFF
--- a/lib/grape-swagger/doc_methods.rb
+++ b/lib/grape-swagger/doc_methods.rb
@@ -47,6 +47,12 @@ module GrapeSwagger
                           'double'
                         when 'Symbol'
                           'string'
+                        when /^\[(?<type>.*)\]$/
+                          items[:type] = Regexp.last_match[:type].downcase
+                          if PRIMITIVE_MAPPINGS.key?(items[:type])
+                            items[:type], items[:format] = PRIMITIVE_MAPPINGS[items[:type]]
+                          end
+                          'array'
                         else
                           @@documentation_class.parse_entity_name(raw_data_type)
                         end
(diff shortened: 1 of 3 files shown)


lib/grape-swagger/doc_methods.rb
```diff
@@ -47,6 +47,12 @@ def parse_params(params, path, method)
                           'double'
                         when 'Symbol'
                           'string'
+                        when /^\[(.*)\]$/
```
dblock: Maybe use a named field here, like type. That avoids having a `last_match[1]` and is probably clearer.

frodrigo:
If it's OK I can squash them.

lib/grape-swagger/doc_methods.rb
```diff
@@ -47,6 +47,12 @@ def parse_params(params, path, method)
                           'double'
                         when 'Symbol'
                           'string'
+                        when /^\[(?<type>.*)\]$/
+                          items[:type] = Regexp.last_match[:type].downcase
+                          if PRIMITIVE_MAPPINGS.key?(items[:type])
+                            items[:type], items[:format] = PRIMITIVE_MAPPINGS[items[:type]]
```
dblock: I re-read this and find it a bit strange (or at least looks like a side-effect) that we're assigning `items[]` inside the case statement. Feels like this logic belongs below where we deal with `items`? I am not sure though, what do you think?

frodrigo: Later we deal with parsed_params[:type], not items[:type], items just affected into parsed_params[:items].

dblock:
Please squash and update CHANGELOG. Thanks.

frodrigo:
Squashed + changelog done

dblock:
Merged.

PR #277, opened 2015-07-17
Removed ruby 1.9.3 and added ruby 2.2.2 to travis.

antek-drzewiecki:
Removed ruby 1.9.3 from travis builds since it is end of life. Added ruby 2.2.2 since its good to be building with a new stable ruby version.


dblock:
Thanks, merged.

Issue #279, opened 2015-07-25
The document example parameter make no sense


birbird:
The code from grape-swagger document

```
params do
  requires :id, type: Integer, desc: 'Coffee ID'
  requires :temperature, type: Integer, desc: 'Temperature of the coffee in celcius', documentation: { example: 72 }
end
```

Does the example field make sense? It will not show on swagger-ui.
Becuase the swagger spec does not have any example definition on parameter object
see http://swagger.io/specification/#parameterObject

dblock:
It's possible that this is just a copy-paste. If you're sure, make a PR that removes it.

birbird:
I created #280 for this issue.

calfzhou:
my documentation.example actually works and i'm now depends on this feature....

mfkp:
I looked this up, I had it in my code and I think I remember it working, but now on the latest version, the example parameter doesn't do anything.

LeFnord:
solved in the actual version

PR #280, opened 2015-07-26
Remove section "Setting a Swagger defaultValue" from readme

birbird:
Swagger does not have a parameter example definition, then the example value does not show on swagger-ui.

### DIFF
--- a/README.md
+++ b/README.md
@@ -205,7 +205,7 @@ desc 'Get a full list of pets', is_array: true
 
 ## Using an options hash
 
-The Grape DSL supports either an options hash or a restricted block to pass settings. Passing the `nickname`, `hidden` and `is_array` options together with response codes is only possible when passing an options hash. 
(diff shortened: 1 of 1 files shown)


dblock:
This was added in https://github.com/tim-vandecasteele/grape-swagger/pull/223 via https://github.com/tim-vandecasteele/grape-swagger/issues/215. So who is right? :)

cc: @jv-dan

jv-dan:
Yeah, I was told using the real swagger property defaultValue was confusing compared to grapes default, but example does populate swagger's defaultValue property without setting a default value on grapes end when the parameter is empty.

dblock:
@birbird @jv-dan I am confused ... someone has to explain this to me like a 2 year old :) 

I want to know what to do with this issue and with #279, please.

dblock:
Adding @calfzhou to this.

@birbird can you please make sense of this?

LeFnord:

PR #284, opened 2015-08-10
make api doc `name` param recognizable by `declared`

calfzhou:
I use a `after_validation` to my api to make sure that all endpoints only get declared params, code sample like:

``` ruby
after_validation do
  @raw_params = params
  @params = declared(params, include_missing: false)
end
```

But this will make swagger doc not available, because this `after_validation` also applies to swagger doc related endpoints, but the required param `:name` cannot be recognized by `declared`.

So i made this PR to make swagger doc endpoints' `:name` param being **declared**.

### DIFF
--- a/lib/grape-swagger/doc_methods.rb
+++ b/lib/grape-swagger/doc_methods.rb
@@ -400,14 +400,11 @@ module GrapeSwagger
         output
       end
 
-      desc specific_api_doc.delete(:desc), { params: {
-        'name' => {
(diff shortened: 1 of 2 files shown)


dblock:
Yes, this is much better and now actually possible. The old syntax is from a very old version of Grape.

dblock:
I merged a bit too quickly, I think a CHANGELOG line is actually useful here, if you can please PR one?

PR #286, opened 2015-08-18
Use `detail` value for `notes`

### LINKED ISSUE #212: How to define `notes` to an operation?
i'm using grape 0.10.0 and grape-swagger 0.9.0.

As mentioned by [grape documentation](https://github.com/intridea/grape#describing-methods), i can add `detail` to each method, but the generated swagger doc doesn't contain information defined in `detail`. Swagger's operation `notes` filed is always an empty string.

I checked grape-swagger source code, the `notes` is generated by the following code:
```ruby
notes       = @@documentation_class.as_markdown(route.route_notes)
```
But i've no idea about what `route.route_notes` is.

rngtng:
Use `detail` value for `notes`

This adds support for using the `detail` value defined within a block passed to `desc` as `notes` value exposed in swagger doc. Next I'd deprecate `notes` and prefer `detail` as specified by grape instead.

CAUTION: only works on grape version >= 0.12.0

```
$ GRAPE_VERSION=HEAD bundle exec rspec ./spec/default_api_spec.rb:62
```

fixes #212

Please have a look at the spec, is this the right format and place to put? I had a hard time to figure out how the specs are structured. pls advise what’s preferred approach..

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,8 @@
 * [#206](https://github.com/tim-vandecasteele/grape-swagger/pull/206): Fixed 'is_array' in the return entity being ignored - [@igormoochnick](https://github.com/igormoochnick).
 * [#266](https://github.com/tim-vandecasteele/grape-swagger/pull/266): Respect primitive mapping on type and format attributes of 1.2 swagger spec - [@frodrigo](https://github.com/frodrigo).
 * [#268](https://github.com/tim-vandecasteele/grape-swagger/pull/268): Fixed handling of `type: Array[...]` - [@frodrigo](https://github.com/frodrigo).
-* [#284](https://github.com/tim-vandecasteele/grape-swagger/pull/284): Use new params syntax for swagger doc endpoint, fix an issue that `:name` params not recognized by `declared` method -[@calfzhou](https://github.com/calfzhou).
+* [#284](https://github.com/tim-vandecasteele/grape-swagger/pull/284): Use new params syntax for swagger doc endpoint, fix an issue that `:name` params not recognized by `declared` method - [@calfzhou](https://github.com/calfzhou).
+* [#286](https://github.com/tim-vandecasteele/grape-swagger/pull/286): Use `detail` value for `notes` - fix an issue where `detail` value specified in a block passed to `desc` was ignored - [@rngtng](https://github.com/rngtng).
 
 ### 0.10.1 (March 11, 2015)
 
--- a/Gemfile
+++ b/Gemfile
@@ -4,7 +4,7 @@ gemspec
 
 case version = ENV['GRAPE_VERSION'] || '~> 0.9.0'
 when 'HEAD'
-  gem 'grape', github: 'intridea/grape'
+  gem 'grape', github: 'ruby-grape/grape'
 else
   gem 'grape', version
 end
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ API class name.
 
 #### markdown
 
-Allow markdown in `notes`, default is `nil`. (disabled) See below for details.
+Allow markdown in `detail`/`notes`, default is `nil`. (disabled) See below for details.
 
 #### hide_format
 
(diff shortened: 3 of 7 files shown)


CHANGELOG.md
```diff
@@ -21,6 +21,7 @@
 * [#266](https://github.com/tim-vandecasteele/grape-swagger/pull/266): Respect primitive mapping on type and format attributes of 1.2 swagger spec - [@frodrigo](https://github.com/frodrigo).
 * [#268](https://github.com/tim-vandecasteele/grape-swagger/pull/268): Fixed handling of `type: Array[...]` - [@frodrigo](https://github.com/frodrigo).
 * [#284](https://github.com/tim-vandecasteele/grape-swagger/pull/284): Use new params syntax for swagger doc endpoint, fix an issue that `:name` params not recognized by `declared` method -[@calfzhou](https://github.com/calfzhou).
+* [#286](https://github.com/tim-vandecasteele/grape-swagger/pull/286): Use `detail` value for `notes` - fix an issue where `detail` value specified in a block passed to `desc` was ignored. -[@rngtng](https://github.com/rngtng).
```
dblock: There's an extra period after `ignored.` :)

I think these are two separate problems, so there should be two separate lines.

dblock: Do we need to update README on this?

rngtng: well, yes a shot note in README would help. I'll add it

rngtng: done, what du you think? I though not to mention `notes:` as its deprecated soon(?) anyway.

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@
 * [#266](https://github.com/tim-vandecasteele/grape-swagger/pull/266): Respect primitive mapping on type and format attributes of 1.2 swagger spec - [@frodrigo](https://github.com/frodrigo).
 * [#268](https://github.com/tim-vandecasteele/grape-swagger/pull/268): Fixed handling of `type: Array[...]` - [@frodrigo](https://github.com/frodrigo).
 * [#284](https://github.com/tim-vandecasteele/grape-swagger/pull/284): Use new params syntax for swagger doc endpoint, fix an issue that `:name` params not recognized by `declared` method -[@calfzhou](https://github.com/calfzhou).
+* [#286](https://github.com/tim-vandecasteele/grape-swagger/pull/286): Use `detail` value for `notes` - fix an issue where `detail` value specified in a block passed to `desc` was ignored -[@rngtng](https://github.com/rngtng).
```
dblock: Now it's missing a space after -, I know it's the same above if you don't mind fixing both? No big deal I can always fix it later.

README.md
```diff
@@ -220,6 +219,16 @@ desc 'Get all kittens!', {
 get '/kittens' do
 ```
 
+## Specify endpoint details
+To specify further details for an endpoint, use the `detail` option within a block passed to `desc`:
```
dblock: Add an empty line above.

README.md
```diff
@@ -220,6 +219,16 @@ desc 'Get all kittens!', {
...
+
+``` ruby
+desc 'Get all kittens!' do
+  detail 'this will expose all the kittens'
+end
```
dblock: So this works just like notes. Why don't we alias one as the other instead?

rngtng: well `notes` is grape-swagger specific, `detail` is supported by grape  in general. so `notes` should be migrated to `detail`. What do you think to write?

dblock: I see, makes sense. Lets leave it at this.

dblock:
I forgot something, I think we need an entry for https://github.com/tim-vandecasteele/grape-swagger/blob/master/UPGRADING.md. If you were using notes before it will get overwritten by detail now, right? That's what we want, but it should be called out to avoid surprises.

dblock:
Next, the build will run against many versions of Grape. It will be failing now, and needs to be fixed somehow, probably with a `pending` depending on the version of Grape.

rngtng:
ok sure, are there any plans to drop support for older grape versions? beeing backwards compatible to the last, let's say, three previous version sounds reasonable to me..

dblock:
In theory yes, in practice not really. I think we should drop support only if there's a good reason. Interestingly this is the *first* feature as far as I know. I would be willing to drop support for something major, like https://github.com/tim-vandecasteele/grape-swagger/issues/219

rngtng:
ok sure fine with me.

rngtng:
ok added minor updates, mainly  README & UPGRADING

Gemfile
```diff
@@ -4,7 +4,7 @@ gemspec
 
 case version = ENV['GRAPE_VERSION'] || '~> 0.9.0'
 when 'HEAD'
-  gem 'grape', github: 'intridea/grape'
+  gem 'grape', github: 'ruby-grape/grape'
```
dblock: Thanks for this.

dblock:
I don't see UPGRADING.md changes :(

rngtng:
sorry, changes are now pushed.

dblock:
Very good, merging.

rngtng:
great, thanks!

calfzhou:
cool!

PR #289, opened 2015-08-25
Grape::Entity >= 0.5.0 compatibility

marshall-lee:
It's failing yet.

Just want to make sure that new `.travis.yml` does the thing.


dblock:
I think you could split refactoring with the version compat (I am ok if you don't), and make sure to update CHANGELOG and such.

marshall-lee:
It is early to add anything to `CHANGELOG`. This thing requires some work in `Grape::Entity` too. I want to improve some things in `documentation` builder.

dblock:
:+1:

lib/grape-swagger/doc_methods/entity_parser/old.rb
```diff
@@ -0,0 +1,71 @@
+module GrapeSwagger
+  module DocMethods
+    module OldEntityParser
+      def parse_entity_models(models)
```
LeFnord: good idea, may be it could adopted for the swagger2.0 branch …

think functionality is implemented, now it would be refactored to have a maintable code, isn't it?
mean, please split it of into some more methods to better dive in, thanks ;)

LeFnord:

Issue #222, opened 2015-03-04
Supporting i18n?


calfzhou:
swagger spec doesn't natively support i18n, but is it possible that grape-swagger generate localized documents for different locale? For example, visiting http://localhost:3000/swagger_doc.json will get default language, as well as visiting http://localhost:3000/swagger_doc.json?locale=zh-CN will get another json in Chinese.

dblock:
It's definitely a useful feature request.

PR #292, opened 2015-08-27
Calf/i18n

calfzhou:
implements #222

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,6 +93,10 @@ API class name.
 
 Allow markdown in `detail`/`notes`, default is `nil`. (disabled) See below for details.
 
+#### i18n_scope
+
+Translations scope (or array of scopes) default is `:api`. [See below for details](#i18n).
+
 #### hide_format
 
 Don't add `.(format)` to the end of URLs, default is `false`.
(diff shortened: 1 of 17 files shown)


calfzhou:
build fail for grape >= 0.10.0, trying to fix it

dblock:
:+1: :+1: :+1:

README.md
```diff
@@ -512,6 +516,208 @@ get '/', http_codes: [
...
+
+```yaml
+api:
+  kittens:
+    get:
+      desc: Get all kittens!
+      detail: this will expose all the kittens
+```
+
+If a translation cannot be found, it will usually fallback to a blank string.
```
dblock: Maybe "usually" is not needed, since it *will* fallback to a blank string. Maybe this:

`If a translation cannot be found, it will use the value specified in the API detail, then default to blank.`

README.md
```diff
@@ -512,6 +516,208 @@ get '/', http_codes: [
...
+
+### Customization
+
+It's very easy to customize the translation process, by set a non-trivial value to the message to be
+translated. As said above, `nil` or blank string message will trigger I18n translation, but it can
+be:
```
dblock: You don't need to convince people that it's easy, just say what it is.

So, "To customize ... "

README.md
```diff
@@ -512,6 +516,208 @@ get '/', http_codes: [
...
+Say if there is model class `User` inherits `Grape::Entity`, and another model `AdminUser` inherits
+`User`, to translate a property named `:email` of `AdminUser`, the lookup keys are:
+
+1. `entities.admin_user.email`
+1. `entities.user.email`
+1. `entities.default.email`
+
+#### All in All
+
+Here is a piece of example translation messages.
```
dblock: This is not great English :) Change "All in All" to just "Example" and you don't need the "here ..." part.

dblock:
I feel like the fallback and the `nil` story is a bit unclear. I think if I do `detail 'Foobar'` and have a translation, the translation should win, but if it cannot be translated I should get `Foobar`. It looks like I cannot get a "default", aka my locale (English) value, which makes it impossible to document the API in one language, then add translation on top.

calfzhou:
> I think if I do detail 'Foobar' and have a translation, the translation should win

So you suggest that even if the message is a normal string, try to look up a translation based on it, am i right?

dblock:
No I am suggesting that it becomes a default value.

calfzhou:
will refine it later :smile_cat:

calfzhou:
updated ~

dblock:
This looks good., I read it carefully. I think `bypass` should be changed to `translate: true/false`, `bypass` is not very english and it's unclear what you're bypassing. What do you think?

README.md
```diff
@@ -512,6 +516,196 @@ get '/', http_codes: [
...
+            name: Resource name
+```
+
+### Customization
+
+The translation can be customized by using different kind/format of message in source code.
+
+* A string or `nil`: It will become the default message when a translation is not defined.
+* A symbol: Looks up translation using the symbol as a key, but will fallback to default key(s).
+* A hash may contain these information:
```
dblock: Change this to say "A hash with the following keys:", "these information" is not very English.

README.md
```diff
@@ -512,6 +516,196 @@ get '/', http_codes: [
...
+  * key: A symbol, defines custom lookup key.
+  * default: A string, defines the default message when translation can not be found.
+  * bypass: A boolean (default is `false`), set to `true` to bypass translation looking up for this
+    message, use the `:default` value or blank string directly.
+  * scope: A symbol, or a string, or an array of them. Used to replace the `i18n_scope` config
+    value for a custom lookup scope.
+  * Any others: Custom variables which will be sent to the translation for interpolation.
```
dblock: Say "Any other key/value will be sent to the translation function for interpolation."

lib/grape-swagger/doc_methods.rb
```diff
@@ -92,11 +124,13 @@ def parse_params(params, path, method)
...
         end
         name          = (value.is_a?(Hash) && value[:full_name]) || param
+        description = translate(description, scope,
+                                i18n_keys.map { |key| :"#{key}.params.#{name}" })
 
         parsed_params = {
           paramType:     param_type,
           name:          name,
-          description:   as_markdown(description),
+          description:   as_markdown(description.presence),
```
dblock: I always found this weird. Ultimately it's just a `nil` or the description, right? Why does it matter here?

dblock:
This is good to merge? Mind squashing the commits?

calfzhou:
squashed :smiley:

dblock:
Merged, great work.

jvortmann:
Hey @calfzhou , I can't manage to make translations work.

There is a route "videos/{uuid}/status and the key "api.videos.:uuid.status.get.desc" but nothing is show either on swagger json or swagger UI.

I can translate  ```I18n.t("api.videos.:uuid.status.get.desc")``` in the console tough.

Am I missing any config?

I'm using grape-swagger (0.25.3).

calfzhou:
I'm not sure, maybe the `:uuid` should be enclosed by quotes in language file?

LeFnord:
please see here: https://github.com/ruby-grape/grape-swagger/issues/465

jvortmann:
@calfzhou it was, but @LeFnord link explains it.

Sorry @LeFnord I tought that issue was about documentation missing and didn't look it.

PR #296, opened 2015-09-04
adding root_prefix to support sub_path apps

pschrammel:
use case: application runs in a sub path of the web server (I was able to configure rails to know about this path with the relative_url_root options). But the generated api paths would still be wrong. This adds a root_prefix for this. Would be nice to get this upstream.

### DIFF
--- a/lib/grape-swagger/doc_methods.rb
+++ b/lib/grape-swagger/doc_methods.rb
@@ -384,17 +384,18 @@ module GrapeSwagger
         info: {},
         authorizations: nil,
         root_base_path: true,
+        root_prefix: '',
         api_documentation: { desc: 'Swagger compatible API description' },
         specific_api_documentation: { desc: 'Swagger compatible API description for specific API' }
       }
 
       options = defaults.merge(options)
-
       target_class     = options[:target_class]
       @@mount_path     = options[:mount_path]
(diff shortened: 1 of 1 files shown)


dblock:
This would need documentation, tests, etc to begin with.

pschrammel:
Am 04.09.2015 um 23:33 schrieb Daniel Doubrovkine (dB.) @dblockdotorg:
> This would need documentation, tests, etc to begin with.
> 
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ruby-grape/grape-swagger/pull/296#issuecomment-137858277>.
> 
Sure. As I don't know the gem that well the first questions on my side
are: Is the code/place correct & would this feature go upstream (test +
doc included).

dblock:
I think this looks file, although I really would like to hear what actual setup requires this. In theory when we mount swagger docs things should be relative and just work.

pschrammel:
My setup: I have the application running in apach with passenger in a subpath http://example.com/app1 http://example.com/app2. I was able to tell rails that it's in a subpath and should generate urls and route everything correctly. The only problem was to tell grape swagger that everything is in a subpath.

dblock:
See above for what's needed for this to be merged.

LeFnord:

PR #297, opened 2015-09-11
Correct use of documentation param_type

fab-girard:
Hello,
when params are defined using an existing grape entity documentation, param_type was not taken into account. I've commited a first modification for that issue.

But the correction has shown another issue : with a param_type now added, and if param is an array, data_type is now set to "array". I think this is inconsitant with allowMultiple parameter... In addition, according to swagger-spec 1.2 allowMultiple is valid only for some param_type ("query", "header" or "path"). Reason of the second commit.

Fab

### DIFF
--- a/README.md
+++ b/README.md
@@ -310,7 +310,7 @@ Grape uses the option `default` to set a default value for optional parameters. 
 
 ## Grape Entities
 
-Add the [grape-entity](https://github.com/agileanimal/grape-entity) gem to our Gemfile.
+Add the [grape-entity](https://github.com/ruby-grape/grape-entity) gem to our Gemfile.
 
 The following example exposes statuses. And exposes statuses documentation adding :type and :desc.
 
--- a/lib/grape-swagger/doc_methods.rb
+++ b/lib/grape-swagger/doc_methods.rb
@@ -103,8 +103,8 @@ module GrapeSwagger
         values               = value.is_a?(Hash) ? value[:values] : nil
         enum_or_range_values = parse_enum_or_range_values(values)
 
-        if value.is_a?(Hash) && value.key?(:documentation) && value[:documentation].key?(:param_type)
-          param_type = value[:documentation][:param_type]
+        if value.is_a?(Hash) && value.key?(:param_type)
+          param_type = value[:param_type]
           if is_array
             items     = { '$ref' => data_type }
             data_type = 'array'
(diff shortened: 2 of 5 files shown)


lib/grape-swagger/doc_methods.rb
```diff
@@ -133,7 +133,7 @@ def parse_params(params, path, method, options = {})
           description:   as_markdown(description),
           type:          data_type,
           required:      required,
-          allowMultiple: is_array
+          allowMultiple: is_array && data_type != 'array' && %w(query header path).include?(param_type)
```
dblock: Maybe make it a variable?

dblock:
Needs a CHANGELOG entry and a build fix, please (rubocop), and squash your commits. Thanks.

lib/grape-swagger/doc_methods.rb
fab-girard: Not sure to understand the benefit if it is just an intermediate variable

fab-girard:
Ok, I've modified CHANGELOG, checked with rubocop and squashed my commit.
Note : a specific test has been commited for each change.

README.md
```diff
@@ -310,7 +310,7 @@ Grape uses the option `default` to set a default value for optional parameters.
 
 ## Grape Entities
 
-Add the [grape-entity](https://github.com/agileanimal/grape-entity) gem to our Gemfile.
+Add the [grape-entity](https://github.com/ruby-grape/grape-entity) gem to our Gemfile.
```
dblock: Thanks for fixing this one.

lib/grape-swagger/doc_methods.rb
```diff
@@ -103,8 +103,8 @@ def parse_params(params, path, method, options = {})
         values               = value.is_a?(Hash) ? value[:values] : nil
         enum_or_range_values = parse_enum_or_range_values(values)
 
-        if value.is_a?(Hash) && value.key?(:documentation) && value[:documentation].key?(:param_type)
-          param_type = value[:documentation][:param_type]
```
dblock: If there's no `documentation` key this will fail, it's clearly a change here that is not caught in tests (and should be probably), right?

dblock: Oh nevermind, I get it.

lib/grape-swagger/doc_methods.rb
dblock: Just more readable otherwise non, nbd.

dblock:
Merging. Thanks.

PR #304, opened 2015-10-15
Remove underscores from default descriptions

epaulet:
This is to prevent descriptions that look like `"Operations about credit_cards"`


dblock:
I'd like not to hack this, but humanize the string using http://api.rubyonrails.org/classes/ActiveSupport/Inflector.html#method-i-humanize or similar.

This also needs tests and a CHANGELOG, please.

LeFnord:

PR #305, opened 2015-10-19
Speedup by parsing models smarter, not harder

jhollinger:
We have one namespace with Very Large entities in its responses, and that swagger doc endpoint began timing out behind our gateway, breaking Swagger UI.

I tracked it down to one line being run in a loop that (seemingly) didn't need to be. Moving it to _after_ the loop brought it down from ~49 sec to ~4 sec on my laptop. Changing the `models` array from an `Array` to a `Set` brought it down to ~3 sec.

Existing tests pass, but I'm not sure of a great way to write one for this since the behavior remains the same. Well, functionally the same. The resulting JSON has the models in a different order, but I don't think that matters.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -8,7 +8,7 @@
 
 # Offense count: 10
 Metrics/AbcSize:
-  Max: 213
+  Max: 214
 
 # Offense count: 1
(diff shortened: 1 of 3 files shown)


dblock:
Very good, merging.

PR #311, opened 2015-10-29
upgrade to produce swagger spec2.0

LeFnord:
these are the first steps, to be compatible to swagger spec 2.0, this introduces also breaking changes with spec 1.2

this isn't ready for now, but the produced swagger.son is compatible with AWS API-Proxy


dblock:
Can you rebase this against master? Thx.

LeFnord:
I can try it, but I think it is impossible, cause of breaking changes, that's why I prefer another branch … give me a day ;)

dblock:
Sorry I am confused, how would a branch make it any different?

LeFnord:
the intention of a second branch is, to have both versions in parallel, and yes they are not mergable

dblock:
I think we don't understand each-other :) First, do everything you need to do on your fork. Make a pull request into `master` here, which I am not going to merge, I will do a `git branch swagger-2.0`, which will make master -> swagger-2.0, then merge into the latter. Right now I cannot do that because your fork is not rebased against master, I'll have to do all the work ;) Makes sense?

LeFnord:
yeap, ok, then I'm closing this PR for now and make my/our work …

salimane:
@LeFnord :+1: 
any update on this ?
Thanks

LeFnord:
@salimane please have a look on the fork,
there is same progress on and for now, one can import the created swagger.json into AWS API proxy and Azure to create/configure the API

salimane:
@LeFnord great job on this :+1: 
Currently what is remaining to merge this back to master ?
I could also help if needed.
Thanks

dblock:
I would love this to move forward! Just need a clean PR with passing specs.

dblock:
I am going to reopen this. Looking for someone to finish it.

LeFnord:
I'm on, but it would be a good idea to have some more eyes on to have a look ;)

dblock:
I am looking at it!

* I think you should put back the paths for specs so we can see the diff clearly. Then we can re-add 1.2 support if we _really_ want to.
* I think you might want to give [swagger-core](https://github.com/swagger-rb/swagger-rb) a try, possibly after this PR, so we can start using a thirdparty library for generating docs.

LeFnord:
@dblock, thanks …
- think it is impossible at the moment to support both spec versions, for this the gem should be configurable, think the easiest solution for the moment would be to have a spec2.0 branch ;)
- your second point sounds reasonable, maybe a direct swagger.{json|ym} file creation could be implemented

dblock:
:+1:

LeFnord:
can anybody give me a hint, how the [built-in validators](https://github.com/ruby-grape/grape#built-in-validators) could be map to the swagger spec 2.0?

LeFnord:
Hi @dblock think it is in a state, where it could be merged into a swagger v2.0 branch :smiley:

the question above (mapping of built-in validators) could/should be solved together with the community, to decide how it should be handled, thanks

dblock:
I have merged this onto https://github.com/ruby-grape/grape-swagger/tree/swagger-2.0. 

I think the next step is to decide whether we're going to make grape-swagger 2.0 only or compatible with multiple specs. Ideally I'd like both, but maybe it's too much work. If we decide to do both, need to bring in the implementation for 1.x back side-by-side. I would start by re-organizing `master` to look more like swagger-2.0 branch's folder structure for swagger 1.x without changing any specs, then do a master -> swagger-2.0 merge so we can continue evolving both side-by-side until we have something that can be released.

LeFnord:
sounds like a plan :smile:

lib/grape-swagger.rb
```diff
@@ -49,6 +40,26 @@ def add_swagger_documentation(options = {})
 
       private
 
+      def version_for(options)
+        options[:version] = version if version
+      end
+
+      def combine_routes(app, doc_klass)
```
florianpilz: You did lots of good simplification in this file and extracting this block to a method is a huge improvement. It still has an enormous complexity - would love to see this method split down further if possible.

lib/grape-swagger.rb
```diff
@@ -111,6 +119,33 @@ def combine_namespace_routes(namespaces)
         end
       end
 
+      def sub_routes_from(parent_route, sub_namespaces)
+        sub_ns_paths = sub_namespaces.collect { |ns_name, _| ["/#{ns_name}", "/:version/#{ns_name}"] }
+        sub_routes = parent_route.reject do |route|
```
timlapluie: `sub_routes` is not needed here, this could just be left out to return the result of `reject`.

lib/grape-swagger.rb
```diff
@@ -111,6 +119,33 @@ def combine_namespace_routes(namespaces)
...
+      def route_instance_variable_equals?(route, name)
+        route_instance_variable(route) == "/#{name}" ||
+          route_instance_variable(route) == "/:version/#{name}"
+      end
+
+      def route_path_start_with?(route, name)
+        route_prefix = route.route_prefix ? "/#{route.route_prefix}/#{name}" : "/#{name}"
+        route_versioned_prefix = route.route_prefix ? "/#{route.route_prefix}/:version/#{name}" : "/:version/#{name}"
+
+        route.route_path.start_with?(route_prefix) ||
```
wteuber: I'd prefer using a regex covering both cases.

lib/grape-swagger.rb
```diff
@@ -49,6 +40,26 @@ def add_swagger_documentation(options = {})
 
       private
 
+      def version_for(options)
+        options[:version] = version if version
+      end
+
+      def combine_routes(app, doc_klass)
+        app.routes.each do |route|
+          route_path = route.route_path
```
thomet: I like that you created a new method for that. But unfortunately it's still hard to understand. Maybe you can generate some more methods to make the code more readable. E.g.:
A separate questionmark method for `route_path.split(/^.*?#{route.route_prefix.to_s}/).last`, `route_match.match('\/([\w|-]*?)[\.\/\(]') || route_match.match('\/([\w|-]*)$')` and `doc_klass.hide_documentation_path && route.route_path.match(/#{doc_klass.mount_path}($|\/|\(\.)/)`

With that, the could would be better readable and the method lengths will increase.

PR #312, opened 2015-10-29
Default values specified with proc fail

maedhr:
Grape allows for default values to be delayed until execution time. This causes the proc itself to be rendered as the Swagger default


dblock:
I am not against this, please fix the build, document this in README, udpate CHANGELOG.

LeFnord:

PR #321, opened 2015-12-24
Prevent crashing when API has camelcase namespaces

gekola:


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 * Your contribution here.
 
+#### Fixes
+
+* [#321](https://github.com/ruby-grape/grape-swagger/pull/321): Fixed handling pathes containing uppercase letters - [@gekola](https://github.com/gekola)
```
dblock: Missing a period at the end :)

dblock: Pathes => paths.

dblock:
I think this makes sense, but I wonder why this was being downcased before, did you track down the origin of that code?

dblock:
Also, what "crashing" does this prevent?

gekola:
With the latest version I get
```.../vendor/bundle/ruby/2.2.0/gems/grape-swagger-0.10.4/lib/grape-swagger.rb:80:in `block in combine_namespace_routes': undefined method `collect' for nil:NilClass (NoMethodError)```
on application load. The namespace was put in downcase to the `@target_class.combined_routes`.

The `downcase!` was introduced in 033bc8013426b774da553375315f1d1ef1c016d0. There are no specific tests added for the case and no comments that indicate it is required by any other code.

dblock:
:+1: merging (the build is failing on rbx, different story)

PR #327, opened 2015-12-30
Re-add Rubinius and set to allow failures on Travis

brixen:
There are many reasons that a build can fail on Travis. In the case of Rubinius, two of the most common have historically been RVM failing to install correctly and general stability issues in Rubinius. However, stability issues are not unique to Rubinius. I have had many builds segfault on MRI as well.

Travis provides the ability to allow failures in selected matrix entries without failing the overall build. This gives visibility without impacting workflow.

Finally, Rubinius development moves very fast and we repeatedly encouraged using 'rbx-2' in .travis.yml to pick up the newest release. When issues are fixed, the new Rubinius release runs and the results are visible.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ rvm:
   - 2.1
   - 2.0
   - jruby-19mode
+  - rbx-2
 
 env:
   - GRAPE_VERSION=0.8.0
@@ -18,3 +19,7 @@ env:
   - GRAPE_VERSION=0.13.0
   - GRAPE_VERSION=0.14.0
   - GRAPE_VERSION=HEAD
+
+matrix:
+  allow_failures:
+    - rvm: rbx-2


dblock:
Thanks @brixen. 

I am not against merging this, but the build doesn't really provide much value, it segfaults at Rubocop, never gets to tests for this project, so we're not really seeing any results.

I also don't want to give people the wrong impression that we support Rubinius, can't make that claim with no tests being run against it, I personally voted to just remove it with https://github.com/ruby-grape/grape-swagger/pull/326, but I would love to hear other people's opinions, please.

brixen:
@dblock

1. I've released 6 versions of Rubinius is the past 3 weeks. If you use 'rbx-2', you'll run against the most recent release. When issues are fixed, you'll see the builds pass. If there is an issue, you'll see the failure, but it won't fail your build. No Ruby implementation will ever not have errors at some point. Saying that this doesn't really provide any value because the build has a failure is very unreasonable.
2. Running Rubocop on every build is a complete waste of power and other resources. Running Rubocop on a single entry in the matrix tells you everything you need from Rubocop. Whatever Rubocop says has zero relevance to whether the code functions. So preventing the specs from running because Rubocop doesn't run is also unreasonable.
3. Running the tests on Rubinius doesn't automatically mean you support Rubinius. Your documentation should make it clear what you support.
4. Whether or not you run on Rubinius may not have any perceived value to you, in which case, don't waste your time. My position would be that the correct consideration is whether it has any value to your users.

dblock:
Merged via https://github.com/ruby-grape/grape-swagger/commit/13666281f82c5dea16d49fcede363e06cd998f73.

Issue #317, opened 2015-12-16
How to hide version from path in swagger


synth:
We are using path versioning in grape. The api is accessible via /api/v2 where v2 is the version.

We would like the swagger documentation to have a base path of /api/v2 and all endpoints to omit this.

From looking at the source, this does not currently seem possible. Is there a best practice here?

From looking at GrapeSwagger::DocMethods#parse_path, it seems it would be an easy place to put a configuration option to exclude the path from the output.

I'm happy to put together a PR if this sounds interesting.

Thoughts?

synth:
I've discovered that this is only an issue when the grape api is structured so that the version is a module such as

````
module Api
  module V2
    class Base < Grape::Api
     mount Api::V2::Thing
     add_swagger_documentation
    end
  end
end
````

synth:
PR: https://github.com/ruby-grape/grape-swagger/pull/319

LeFnord:
for the actual, see: [add_version](https://github.com/ruby-grape/grape-swagger#add_version)

PR #330, opened 2016-01-06
Hide version from path in swagger

synth:
#317 - hiding version module from path

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ### 0.10.5 (Next)
 
-* Your contribution here.
+-* Your contribution here.
+* [#330](https://github.com/ruby-grape/grape-swagger/pull/319): Hide module version from path [@synth](https://github.com/synth).
 
 #### Fixes
 
--- a/lib/grape-swagger/doc_methods.rb
+++ b/lib/grape-swagger/doc_methods.rb
@@ -208,7 +208,7 @@ module GrapeSwagger
       end
     end
 
-    def parse_path(path, version)
+    def parse_path(path, version, hide_module_from_path = false)
       # adapt format to swagger format
       parsed_path = path.sub(/\(\..*\)$/, @@hide_format ? '' : '.{format}')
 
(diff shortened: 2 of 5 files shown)


synth:
@dblock 3rd PR's a charm!

lib/grape-swagger/doc_methods.rb
```diff
@@ -491,7 +497,7 @@ def setup(options)
         end
 
         ops = visible_ops.group_by do |route|
-          @@documentation_class.parse_path(route.route_path, api_version)
+          @@documentation_class.parse_path(route.route_path, api_version, hide_module_from_path)
```
dblock: Why is this passed in and not assigned as a global `@@hide_module_from_path` like everywhere else?

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ### 0.10.5 (Next)
 
-* Your contribution here.
+-* Your contribution here.
```
dblock: This is still busted.

screen shot 2016-01-06 at 4 25 54 pm

dblock:
You should learn how to amend the change in git, next time you make a change, do `git commit --amend`, which will edit the previous commit, then force push the branch with `git push origin hide-module-from-path -f`, this way you won't have to redo the PR every time.

lib/grape-swagger/doc_methods.rb
synth: I initially did it that way, but had some problem I can't remember. When I fix up again, I"ll try with class var and see if tests pass

synth: @dblock So, using class variables precludes the ability to define multiple apps with different swagger doc options as the class variables are defined internally on the DocMethods module(the one and only) rather than the anonymous documentation class that is created in `add_swagger_documentation` implementation. My tests create two apps with different options but fail with the class variable approach because it gets clobbered whatever is the value of the 2nd app. This is obviously a bigger issue than my one little new option but that happens to be caught by the way I wrote my test. What are your thoughts?

dblock: That is a definite problem and feel free to fix it up in a different PR. Want to do that first and we'll merge this after?

My biggest gripe is introducing N parameters here.

I think this like should be:

```ruby
documentation_class.parse_path(options.merge(path: route.route_path, version: api_version))
```

synth: It would be cool to have them as class vars or class instance vars on each dynamically created documentation class rather than the global module and rather than needing to pass options around all over the place. Ill see if I can come up with a nice approach.

PR #336, opened 2016-01-28
upgrades to grape 0.14.x; grape-entity 0.5.x

LeFnord:


dblock:
Thanks. Fix CI please (`rubocop -a`), squash commits.

LeFnord:
sorry I have not the time at the moment to fix it, I pay more attention on a green test suite,
as one could see here [grape-swagger cadetship](https://codeship.com/projects/104499/builds/11729901) and adding more features

LeFnord:
please merge so that I can work on,

is it possible to cut off my fork and branch from this one, so that I can work on my fork
and make a PR to yours independent from my PRs of feature branches?
so travis don't need to run on every commit, thanks

the failed builds are rubinius, could not find a binary of it
and jruby, has problems with utf-8 chars, but this seems to be a standard java problem, not ruby

spec/swagger_v2/api_swagger_v2_format-content_type_spec.rb
```diff
@@ -0,0 +1,104 @@
+require 'spec_helper'
+
+describe 'format, content_type' do
+  include_context "the api entities"
+
+  before :all do
+    module TheApi
```
dblock: This is fine, however these classes will be declared globally, you might want to change it to a dynamic class, like `Class.new(Grape::API)` or namespace them with something closer to the spec name, like `module ContentTypeSpec ...`.

dblock:
This is great, lets try to finish it.

* We don't need to worry about Rubinius, replace `rbx-2.2.10` with `rbx-2` and add it to `allow_failures`. 
* You do need to fix the JRuby build, the problematic file is most likely just missing a UTF-8 marker on top.
* Please update CHANGELOG and README. 
* Squash your commits.
* Is this still compatible with grape-entity 0.4.x? What else is changing that needs to be in UPGRADING?

LeFnord:
thanks for the response, I'll try to do it in the next hours …

to your questions:

- yes it is compatible with grape-entity 0.4.8 and 0.5.0, it was a requirement
- the README should be up-to-date, have to do some work on the CHANGELOG

a last question: can I change the version to something starting with 0.2.x, to indicate, breaking changes,
and create a tag for that (this is required from my firm, to use it)

dblock:
Next version should be 0.11.0, you can update that. Don't worry about tagging, I'll cut a release once we have everything on green.

dblock:
This is looking good. Should I merge this? Squash commits? Update CHANGELOG?

LeFnord:
you can merge it, if you want :smile: 
the changeling is updated and I squashed some commit, sorry not all

for the next work, I want to merge this work into my fork (there it is master),
and make a regular PR, is it ok to do so?

next work should be:
- make params related stuff modular
- make adding of user defined stuff available → support for something like `x-foo- …`, first for the paths section, later for definition section

is it ok for you?

dblock:
I merged this one. 

I am down with a PR into master that looks like it could actually ship. The biggest concerns for me are UPGRADING, README and CHANGELOG, where people know what they are getting.

LeFnord:
thanks for merging
no, I don't mean into into this master,
I mean, I want work on 'my' master and make regular PRs to the this swagger-2.0 branch

dblock:
It's your master :)

PR #345, opened 2016-03-03
Moving documented headers into the `parameters` object

daveWid:
The documented headers in the description block describing the request headers. To have them picked up in the SwaggerUI they need to be in the parameters object with in: 'header' instead of a headers object, which used to document the response headers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ n.n.n / 2016-02-05
 
 ### 0.10.3 (Next)
 
+* [#345](https://github.com/ruby-grape/grape-swagger/pull/345): Moving documented headers into the `parameters` object - [@daveWid](https://github.com/daveWid).
+
 [#336](https://github.com/ruby-grape/grape-swagger/pull/336) changes of swagger-2.0 fork, to support it
 
 * updates gems, corrects parameter, which is in array, make rubocop happy
--- a/lib/grape-swagger/endpoint.rb
+++ b/lib/grape-swagger/endpoint.rb
@@ -127,11 +127,12 @@ module Grape
     def method_object(route, options)
       methods = {}
       methods[:description] = description_object(route, options[:markdown])
-      methods[:headers] = route.route_headers if route.route_headers
 
       methods[:produces] = produces_object(route, options)
 
       methods[:parameters] = params_object(route)
+      methods[:parameters].concat(parse_header_params(route)) if route.route_headers
+
       methods[:responses] = response_object(route)
 
       if route.route_aws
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -2,7 +2,11 @@
 n.n.n / 2016-02-05
 ==================
 
-### 0.10.3 (Next)
+### Next Release
+
+* [#345](https://github.com/ruby-grape/grape-swagger/pull/345): Moving documented headers into the `parameters` object - [@daveWid](https://github.com/daveWid).
+
+### 0.10.3
```
dblock: Hasn't shipped, so undo this change please?

lib/grape-swagger/endpoint.rb
```diff
@@ -127,11 +127,12 @@ def path_item(routes, options)
     def method_object(route, options)
       methods = {}
       methods[:description] = description_object(route, options[:markdown])
-      methods[:headers] = route.route_headers if route.route_headers
 
       methods[:produces] = produces_object(route, options)
 
       methods[:parameters] = params_object(route)
+      methods[:parameters].concat(parse_header_params(route)) unless route.route_headers.nil?
```
dblock: The `unless` seems backwards, what was wrong with the `if`?

dblock:
This is good, see minor comments above and please squash your commits.

LeFnord:
Hi @daveWid … think we have to distinguish between the [swagger-header](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#headers-object) and parameters in headers, as you mean.

for that I would propose, that all stuff that is described inside the `params` block is request related
and all from the `desc` is more general stuff

LeFnord:
please have a look here: https://github.com/ruby-grape/grape-swagger/pull/362

songlipeng2003:
@LeFnord I find swagger 2.0 don't support headers in path, so method headers can be response headers for 2xx response, or merge headers to parameters of path .
Or you can will remove headers feature in method, but I use old grape-swagger version, I update it,
and headers params  are lost.
I think that move headers to parameters of path is better.

LeFnord:
sorry @songlipeng2003 it is a bit confusing to me what you want …

set headers on route level is done via `headers`, see: [swagger-header-parameters](https://github.com/ruby-grape/grape-swagger#swagger-header-parameters-)

specifying headers as request parameters is done via `param_type`, see: [api_swagger_v2_param_type_spec.rb](https://github.com/ruby-grape/grape-swagger/blob/master/spec/swagger_v2/api_swagger_v2_param_type_spec.rb)

PR #346, opened 2016-03-04
Fix: Allow is_array for nested entities with with documentation

daveallie:
Previously an endpoint which returned the `MyEntity` entity (below):

```ruby
module Entities
  class SubEntity < Grape::Entity
    expose :some_value, documentation: { type: "String" }
  end
  class MyEntity < Grape::Entity
    expose :sub_entities, using: Entities::SubEntity, documentation: { is_array: true }
  end
end
```

Would generate this documentation:

```javascript
{
  ...
  definitions: {
    MyEntity: {
      type: "object",
      properties: {
        sub_entities: {
          $ref: "#/definitions/SubEntity"
        }
      }
    },
    SubEntity: {
      type: "object",
      properties: {
        some_value: {
          type: "string"
        }
      }
    }
  }
}
```

With the changes it will now produce:

```javascript
{
  ...
  definitions: {
    MyEntity: {
      type: "object",
      properties: {
        sub_entities: {
          type: "array",
          items: {
            $ref: "#/definitions/SubEntity"
          }
        }
      }
    },
    SubEntity: {
      type: "object",
      properties: {
        some_value: {
          type: "string"
        }
      }
    }
  }
}
```

Which better represents the structure and follows the same output standard as if `in_array` was specified on the actual endpoint.

### DIFF
--- a/lib/grape-swagger/endpoint.rb
+++ b/lib/grape-swagger/endpoint.rb
@@ -245,7 +245,11 @@ module Grape
         x[0] = x.last[:as] if x.last[:as]
         if x.last[:using].present? || could_it_be_a_model?(x.last)
           name = expose_params_from_model(x.last[:using] || x.last[:type])
-          memo[x.first] = { '$ref' => "#/definitions/#{name}" }
+          memo[x.first] = if x.last[:documentation] && x.last[:documentation][:is_array]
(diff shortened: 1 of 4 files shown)


dblock:
Merged, thanks.

LeFnord:
:+1:

PR #347, opened 2016-03-07
Fix typeref

c910335:
Fixed: empty type reference when both :using and :as are provided in exposure but no :type in documentation


dblock:
This needs a test and a CHANGELOG entry please.

dblock:
:+1: Github doesn't send a notification after a new push, so just seeing this.

dblock:
@LeFnord check this in swagger-2.0

LeFnord:
yeap, will have an eye on …

PR #353, opened 2016-03-15
fixes issue-352

LeFnord:


dblock:
Take the habit for writing tests for regressions!

LeFnord:
yeap … would be better to work with :+1: and not merging it self …

PR #356, opened 2016-03-16
some improvements

LeFnord:
* [x] fixes setting of `base_path`, `host`
* [x] adds possibility to configure the setting of `version` and `base_path` in documented path
* [x] adds `operationId`
* [x] adds `consumes` setting
* [x] refactoring


dblock:
@LeFnord Btw, trust you to do the right things on swagger-2.0 until we actually ship a release, then we can start being more diligent with PRs, self-merges and what not. I read through this change, excellent work.

Issue #364, opened 2016-04-01
Grape: route_xxx methods have been deprecated


ngonzalez:
Grape Route model route_* methods have been deprecated
https://github.com/ruby-grape/grape/blob/master/lib/grape/router/route.rb#L100

Grape-Swagger still uses a lot of these methods, and the deprecation warnings can be annoying ..

LeFnord:
yeap, this breaking change is known,
as I mentioned in the [grape group](https://groups.google.com/forum/#!topic/ruby-grape/kdP1hTSILLw)
the adoption of is planned for 0.21.0 release

ngonzalez:
@LeFnord Thanks

dblock:
Do we want a 0.10.6 or 0.11.0 that locks a higher version of Grape and fixes the deprecations?

LeFnord:
@dblock … think we should use 0.11.0 for grape >= 0.15.0 to indicate breaking change

dblock:
Reopening this for a 0.11.0. It's a little more involved than I thought.

SuperMasterBlasterLaser:
Is there any Pull requests that fixes this?

pamit:
This may fix this problem,

gem 'grape', '>= 0.16.2'
gem 'grape-entity', '< 0.5.0'
gem 'grape-swagger', '0.11.0'
gem 'grape-swagger-rails'#,  '~> 0.0.8'

It works for me.

dblock:
I'll close this, we have released compatible versions at this point. Understand that the matrix is a little complicated, but it gets better as we move forward. See https://github.com/ruby-grape/grape-swagger#version.

PR #368, opened 2016-04-04
Require Grape 0.16.1 and fix route_xxx deprecations.

dblock:
This isn't ready to merge. there're issues with `route_params` and `route_hidden`.

Closes https://github.com/ruby-grape/grape-swagger/issues/364.


dblock:
For now I am monkey-patching the deprecation to avoid too many warnings being spewed.

```ruby
module Grape
  class Router
    class Route
      private

      # TODO: remove when https://github.com/ruby-grape/grape-swagger/pull/368 and https://github.com/newrelic/rpm/pull/235 are finished
      def warn_route_methods(name, location, expected = nil)
      end
    end
  end
end
```

lib/grape-swagger/doc_methods.rb
```diff
@@ -501,51 +504,57 @@ def setup(options)
...
+
             if operation[:parameters].any? { |param| param[:type] == 'File' }
               operation.merge!(consumes: ['multipart/form-data'])
             end
             operation.merge!(responseMessages: http_codes) unless http_codes.empty?
 
-            if route.route_entity
-              type = @@documentation_class.parse_entity_name(Array(route.route_entity).first)
-              if route.route_is_array
+            if route.entity
```
LeFnord: please respect also `route.success` → see: [describing-methods](https://github.com/ruby-grape/grape#describing-methods)

dblock: that needs a spec I suspect, wasn't caught

LeFnord: `success` is the new `entity` and `failures` the new `http_codes`, think a small spec to proof both new work in the same way as the old one should be enough

.travis.yml
```diff
@@ -10,14 +10,7 @@ rvm:
...
 env:
-  - GRAPE_VERSION=0.8.0
-  - GRAPE_VERSION=0.9.0
-  - GRAPE_VERSION=0.10.0
-  - GRAPE_VERSION=0.10.1
-  - GRAPE_VERSION=0.11.0
-  - GRAPE_VERSION=0.12.0
-  - GRAPE_VERSION=0.13.0
-  - GRAPE_VERSION=0.14.0
+  - GRAPE_VERSION=0.16.1
```
LeFnord: could you add `GRAPE_VERSION=0.15.0`

dblock:
@LeFnord once this is all fixed here, should I be merging it onto master or swagger-1.2?

LeFnord:
@dblock merg to master, after this is done, I would release the new versions

dblock:
Closing in favor of https://github.com/ruby-grape/grape-swagger/pull/375.

lib/grape-swagger/doc_methods.rb
dblock: Actually no need, these were declared as DSL sugar on top of the `entity` and `http_codes` implementation in https://github.com/ruby-grape/grape/blob/1097eb192a8601234b5465c23b1ab09101048792/lib/grape/dsl/desc.rb#L90, so they work today out of the box.

PR #377, opened 2016-04-13
Fix: un-pending-ify spec.

dblock:
Was marked pending in https://github.com/ruby-grape/grape-swagger/commit/55ceec689d5df6506704e6c04481148fbb45e404, but doesn't need to be.


LeFnord:
why is green here?

dblock:
Does it fail for you @LeFnord? Passes on my machine. I don't see why it would either.

LeFnord:
yeap, it fails, also on travis, this was the reason for pending …

dblock:
Well Travis says otherwise in this pull request ;)

dblock:
Maybe order of execution?

LeFnord:
however it works :)

PR #381, opened 2016-04-13
adding entity property description when documentation desc option is present

elciok:
The swagger_doc json generated contains entity definitions, with property declarations, but if you set a "desc" option like this:

expose :name, documentation: {type: "String", desc: "Name of object"}

This gets ignored by grape-swagger. I've changed this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 #### Features
 
+[#381](https://github.com/ruby-grape/grape-swagger/pull/381) 
+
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -35,6 +35,10 @@
 
 [#353](https://github.com/ruby-grape/grape-swagger/pull/353) resolves issue #352
 
+[](https://github.com/ruby-grape/grape-swagger/pull/)
```
LeFnord: please move to the top

LeFnord:
:+1:

PR #382, opened 2016-04-14
Make default schemes undefined to match Swagger Spec

wleeper:
#### What's this PR do, and where should I start?

This PR takes out the default value for schemes.  This brings the generator in line with Swagger Spec 2.0 which indicates that schemes is an optional parameter and default behavior is to assume the same scheme as the scheme of the document.

#### How should this be manually tested?

Can be tested using the example (which did not work prior to this change) by running:

```
bundle exec rackup
```

From the examples directory.  This will start the example up on http://localhost:9292.  Drop the `http://localhost:9292/swagger_doc` into `http://petstore.swagger.io` and choose one of the methods to 'try'. It should successfully hit the back end.

#### Any background context you want to provide?

Ran into this problem because we run our app in development on http only and the default value of `['https', 'http']` failed to work.  It also failed as noted to work with the example if started per the README.

* Tests were updated to reflect the lack of schemes tag in the default.  There was already code to override this value with custom options, those tests were not changed and still pass as expected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+### (next)
+
+#### Fixes
+* [#382](https://github.com/ruby-grape/grape-swagger/pull/382): Make the schemes undefined by default to match Swagger 2.0 spec indicating that a blank schemes will use the scheme from the document - [@wleeper](https://github.com/wleeper).
+
 ### 0.20.0 / 2016-04-09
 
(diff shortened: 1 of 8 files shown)


CraigCottingham:
+1 for an awesome commit message.

LeFnord:
:+1: please update CHANGELOG

PR #383, opened 2016-04-15
fixes grape to 0.14.0

LeFnord:
fixing of grape-entity is not needed, tested it with: 0.4.8, 0.5.0 and 0.5.1 runs green always

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,10 +1,3 @@
 source 'http://rubygems.org'
 
 gemspec
-
-case version = ENV['GRAPE_VERSION'] || '~> 0.14.0'
-when 'HEAD'
-  gem 'grape', github: 'ruby-grape/grape'
-else
-  gem 'grape', version
-end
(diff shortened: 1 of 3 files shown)


Gemfile
```diff
@@ -1,10 +1,3 @@
 source 'http://rubygems.org'
 
 gemspec
-
-case version = ENV['GRAPE_VERSION'] || '~> 0.14.0'
-when 'HEAD'
-  gem 'grape', github: 'ruby-grape/grape'
-else
-  gem 'grape', version
-end
```
dblock: Removing this caused Travis to not build different versions!

https://github.com/ruby-grape/grape-swagger/blob/master/.travis.yml

LeFnord: mmh, then I should read it …

dblock: I also think we should not spend too much time supporting previous releases of things with newer versions, I suggest moving onward and enabling 0.16.x + whatever new stuff over older things.

LeFnord: yeap, I totally agree with you …
but we should support grape 0.14.0 for a while, as I said, support for > 0.16.1 is on the road,
but we should give all a while to test actual grape-swagger version and improve it

Issue #391, opened 2016-04-19
Header is not displaying in Swagger UI


brahamshakti:
This Code is not displaying header field in swagger UI

    desc "Return super-secret information", {
      headers: {
        "XAuthToken" => {
          description: "Valdates your identity",
          required: true
        },
       "XOptionalHeader" => {
         description: "Not really needed",
         required: false
       }
     }
    }

Earlier It used to be like 
![xauthtoken](https://cloud.githubusercontent.com/assets/2370848/14633863/2ae91a4c-063d-11e6-8ec5-d92aa39750ef.png)

But now it is showing no header field with above Code

![noxauthtoken](https://cloud.githubusercontent.com/assets/2370848/14633880/3f2073ac-063d-11e6-871d-39dcc3306845.png)

How do I fix this issue.

LeFnord:
would be cool to write a spec (similar to this one: [api_swagger_v2_headers_spec.rb](https://github.com/ruby-grape/grape-swagger/blob/master/spec/swagger_v2/api_swagger_v2_headers_spec.rb)

so it could be reproduced

brahamshakti:
I am using version 0.20.1.

LeFnord:
ok, I see …
the so defined header should go into the response object,
to define header parameters, take a parameter and set the param_type to header

brahamshakti:
It would be better if you could provide an alternative solution in form of some patch. Sorry but I think modifying 100s of apis to add just one extra parameter does not look a good approach. Can I accept some sort of patch in near future from you

dblock:
@brahamshakti What exactly are you proposing? What would you like to write in your API? We'll accept patches from you, thanks.

wleeper:
doing from the referenced test case

```
        params do
          optional :in_query, type: String, documentation: { param_type: 'query' }
          optional :in_header, type: String, documentation: { param_type: 'header' }
        end
```

does not really work.  It will make valid swagger spec, but the grape code is looking for the `in_header` parameter in the params list not in the header.  If it were a required field it would fail if you actually put the parameter in the header.

LeFnord:
but the question is always, how to differ between stuff for request and stuff for response for documentation …

what are your thoughts?

LeFnord:
@brahamshakti give it a try, and close the issue

brahamshakti:
Thanks to @wleeper, my issue is resolved with this pull request. I am closing this issue for now. Thanks to all.

PR #393, opened 2016-04-22
Added feature to report header params in params block

wleeper:
#### What's this PR do, and where should I start?

This PR address [Issue #391](https://github.com/ruby-grape/grape-swagger/issues/391)

This PR changes the behavior by no longer output a header tag at the root level of an endpoint, and replaces that with parameters tagged with `"in": "header"`, which is the correct Swagger 2.0 method.

See changes to spec/swagger_v2/simple_mounted_api_spec.rb to see how the behavior changed.

#### How should this be manually tested?

Fire up the example app and add some header variables using the proper grape syntax (note the example below is <= 0.14, not what is in the docs for 0.16.2:

```
        desc 'This returns something',
          failure: [{code: 400, model: Entities::ApiError}],
          headers:  {
            "X-Rate-Limit-Limit" => {
              "description" => "The number of allowed requests in the current period",
              "type" => "integer"
          }}
        get '/use_headers' do
          { "declared_params" => declared(params) }
        end
```

#### Any background context you want to provide?

Ran into this issue when I tried to add header params to our app.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 #### Fixes
 
 * [#389](https://github.com/ruby-grape/grape-swagger/pull/389): respect X-Forwarded-Host - [@edvakf](https://github.com/edvakf).
+* [#393](https://github.com/ruby-grape/grape-swagger/pull/393): properly handle header parameters- [@wleeper](https://github.com/wleeper).
 
 ### 0.20.1 / 2016-04-17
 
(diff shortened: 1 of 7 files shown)


LeFnord:
@wleeper thanks for it, but to be consistent with your own proposal, I suggest a stronger distinction between request and response

1. request related: all what is denied inside the params block
2. response related: all from desc section

what did mean?

LeFnord:
👍 but the question is, how to differ between stuff for request and stuff for response …

what are your thoughts?

wleeper:
This just addresses the request parameters.  I'm not even sure exactly where swagger puts response headers and how we would capture the documentation.

LeFnord:
yeap, I know … swagger puts the response headers into the [response object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#response-object),
but how can we document it?

the header key from desc is now a parameter, that was the intention for my proposal above

Issue #395, opened 2016-04-22
Parameter description of null raises a validation error in Swagger-UI


wleeper:
If you don't provide a desc: field for a parameter, the generated swagger spec has `null` as the value for description.  This raises a validation error in Swagger-UI

ex:
```
{
"in": "query",
"name": "force_refresh",
"description": null,
"type": "string",
"required": false
},
```

Produces the following validation error

```
schemaValidationMessages": [
{
"level": "error",
"domain": "validation",
"keyword": "oneOf",
"message": "instance failed to match exactly one schema (matched 0 out of 2)",
"schema": {
"loadingURI": "http://swagger.io/v2/schema.json#",
"pointer": "/definitions/parametersList/items"
},
"instance": {
"pointer": "/paths/test/get/parameters/3"
}
},
```

**Recommendation**
Based on http://swagger.io/specification/#parameterObject the description is optional and should be omitted when nil

LeFnord:
@wleeper have a look on https://github.com/ruby-grape/grape-swagger/pull/399 and give it a try

PR #399, opened 2016-04-24
makes param description optional

LeFnord:
solves: https://github.com/ruby-grape/grape-swagger/issues/395


SuperMasterBlasterLaser:
:+1:

PR #408, opened 2016-05-06
Added head method type

Bugagazavr:
Grape swagger fails to generate doc if head route exists

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 #### Fixes
 
+* [#408](https://github.com/ruby-grape/grape-swagger/pull/408): fix fails when point with head method present - [@Bugagazavr](https://github.com/Bugagazavr)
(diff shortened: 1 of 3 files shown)


dblock:

Bugagazavr:
@dblock ok, when i should place spec?

dblock:
You mean where? I am not sure. You'll have to find a good place, there has been a lot of changes on master I am not too familiar with the updated structure.

Bugagazavr:
done

dblock:
Merged. Is it the same problem for OPTIONS? Want to take care of it?

Bugagazavr:
yep, but this problem can be sovled, by nil check before https://github.com/ruby-grape/grape-swagger/blob/master/lib/grape-swagger/endpoint.rb#L164

dblock:
@Bugagazavr write some specs, make some PRs! Much appreciated.

PR #415, opened 2016-05-10
Fix RuboCop offences

lest:
RuboCop 0.40 includes new cops.

https://github.com/bbatsov/rubocop/releases/tag/v0.40.0


dblock:
We should be locking down the version of Rubocop used, so not to have side effects like this. Would you mind making that change please, I am happy with a Rubocop upgrade.

dblock:
I went ahead and fixed this up in https://github.com/ruby-grape/grape-swagger/commit/76025377bdde28c61120dda08b5d0edd7ad42e1d. Thank you!

lest:
👍 Thanks!

PR #422, opened 2016-05-15
re-adds jruby-9.0.5.0 to travis

LeFnord:


dblock:
Do we really need all these combinations? I would just add one build without all these external dependencies.

LeFnord:
agree, I would only add the latest for jruby

PR #423, opened 2016-05-16
Prepare to new version

Bugagazavr:


LeFnord:
mmh, too late, would be nice to have this https://github.com/ruby-grape/grape-swagger/pull/422 in the new release

PR #428, opened 2016-05-18
Fix broken link in readme

hedgesky:
File {{example/api.rb}} was removed in cc54c86d, so update link in readme to the new location.


LeFnord:
👍

PR #431, opened 2016-05-20
Add fixed fields

aschuster3:
Between Swagger v1.2 and v2, the `description` took a slightly different meaning and `summary` took it's place. I had to re-work a handful of tests so that summary was accepted and in order to be somewhat forward compatible, I made the string passed to `desc` on paths fill in the value for the summary by default with the option to override it.

While I was at it, I also added `deprecated` and built tests around that logic.

I based my additions on [this resource](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#operation-object)

I tested and confirmed that these changes reflect on SwaggerUI appropriately.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,11 +2,9 @@
 
 #### Features
 
-* Your contribution here.
+* [#431](https://github.com/ruby-grape/grape-swagger/pull/431): Add summary and deprecated to the operation object generator to be more compliant with [OpenAPI v2](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#operation-object) - [@aschuster3](https://github.com/aschuster3).
 
 #### Fixes
-
-* Your contribution here.
 
 ### 0.20.4 (May 16, 2016)
 
--- a/README.md
+++ b/README.md
@@ -442,6 +442,52 @@ end
 ```
 
 
+#### Overriding the route summary
+
+By default, the route summary is filled with the value supplied to `desc`.
+
+```ruby
+namespace 'order' do
+  desc 'This will be your summary'
+  get :order_id do
+    ...
+  end
+end
+```
+
+To override the summary, add `summary: '[string]'` after the description.
+
+```ruby
+namespace 'order' do
+  desc 'This will be your summary',
+    summary: 'Now this is your summary!'
+  get :order_id do
+    ...
+  end
+end
+```
+
+
+#### Deprecate routes
+
+For route deprecation, add `deprecated: true` after your description.
+
+```ruby
+namespace 'order' do
+  desc 'Old way to do things',
+    deprecated: true
+  get :old do
+    ...
+  end
+
+  desc 'New way to do things'
+  get :new do
+    ...
+  end
+end
+```
+
+
 #### Expose nested namespace as standalone route
 
 Use the `nested: false` property in the `swagger` option to make nested namespaces appear as standalone resources.
--- a/lib/grape-swagger/endpoint.rb
+++ b/lib/grape-swagger/endpoint.rb
@@ -102,7 +102,9 @@ module Grape
     end
 
     def method_object(route, options, path)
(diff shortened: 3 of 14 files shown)


README.md
```diff
@@ -442,6 +442,50 @@ end
...
+    ...
+  end
+end
+```
+
+To override the summary, add `summary: '[string]'` after the description.
+
+```ruby
+namespace 'order' do
+  desc 'This will be your summary', summary: 'Now this is your summary!'
```
LeFnord: please move `summary`to new line and intend it, for better finding

README.md
```diff
@@ -442,6 +442,50 @@ end
...
+```
+
+
+#### Deprecate routes
+
+For route deprecation, add `deprecated: true` after your description.
+
+```ruby
+namespace 'order' do
+  desc 'Old way to do things', deprecated: true
```
LeFnord: same as above

lib/grape-swagger/endpoint.rb
```diff
@@ -103,18 +103,28 @@ def path_item(routes, options)
...
+      method[:deprecated]  = deprecated_object(route)
       method[:operationId] = GrapeSwagger::DocMethods::OperationId.build(route, path)
       method.delete_if { |_, value| value.blank? }
 
       [route.request_method.downcase.to_sym, method]
     end
 
+    def summary_object(route)
+      summary = route.options[:desc] if route.options.key?(:desc)
+      summary = route.description if route.description.present?
```
LeFnord: this assumes, that, if `summary` not given it is the same as description,
I ask me if it is intended by the spec …

CHANGELOG.md
```diff
@@ -2,11 +2,7 @@
 
 #### Features
 
-* Your contribution here.
-
-#### Fixes
-
-* Your contribution here.
```
dblock: Put this back please.

CHANGELOG.md
```diff
@@ -2,11 +2,7 @@
...
+* [#431](https://github.com/ruby-grape/grape-swagger/pull/431)Add summary and deprecated to the operation object generator to be more compliant with [OpenAPI v2](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#operation-object) - [@aschuster3](https://github.com/aschuster3)
```
dblock: Please make this the same syntax as everything else. Missing `:` after the issue number and a period at the end.

lib/grape-swagger/endpoint.rb
aschuster3: This part was mostly interpretation, but when comparing [Open API v1.2 Operation Object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/1.2.md#523-operation-object) and [Open API v2 Operation Object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#operation-object) specs, `summary` and `notes` are the equivalent to `summary` and `description`, so when upgrading from v1.2 to v2 I would expect:

```ruby
desc 'Example summary' do
    detail 'More details'
end
```

To produce:
```json
{
  ...
  "summary": "Example summary",
  "description": "Example summary\n More details",
  ...
}
```

More importantly, I base on what appears in SwaggerUI so that the route summary appears next to the route name, as pictured:

Pet Store Summary

This was the default behavior for when I used `grape-swagger` version <= 0.11 (something about SwaggerUI may have changed as well), but I'm open to discussion.

LeFnord: I'm also open for proposals 😉 

want say I'm fine with your suggestion above,
maybe you could split of the PR into the `deprecation` stuff and `summary`/`description`…

I want only one point mention, yeap swagger-ui is changing a lot at the moment,
that's why it should be used for view stuff, but not for checking of validity of the generated swagger,
in normal cases, one should prefer http://bigstickcarpet.com/swagger-parser/www/index.html, over it, or one should use [ngrok](https://ngrok.com) (or similar), then this would also be validated

cause IMO a valid swagger is much more important then a nice looking one

aschuster3: Alright, I will split this PR into two separate ones to keep the changes atomic.  I will close this one and open another.

PR #434, opened 2016-05-20
Add summary field

aschuster3:
A redo of PR #431, but with just `summary`.  I will add `deprecated` later.

Between Swagger v1.2 and v2, the `description` took a slightly different meaning and `summary` took it's place. I had to re-work a handful of tests so that summary was accepted and in order to be somewhat forward compatible, I made the string passed to `desc` on paths fill in the value for the summary by default with the option to override it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 #### Features
 
+* [#434](https://github.com/ruby-grape/grape-swagger/pull/434): Add summary to the operation object generator to be more compliant with [OpenAPI v2](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#operation-object) - [@aschuster3](https://github.com/aschuster3).
 * Your contribution here.
 
 #### Fixes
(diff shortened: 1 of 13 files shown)


LeFnord:
👍

CHANGELOG.md
```diff
@@ -2,12 +2,10 @@
 
 #### Features
 
-* Your contribution here.
```
dblock: Put this back please!

lib/grape-swagger/endpoint.rb
```diff
@@ -102,7 +102,9 @@ def path_item(routes, options)
     end
 
     def method_object(route, options, path)
+      puts "lib/grape-swagger/endpoint.rb:105 DEBUG OUTPUT FOR GRAPE-SWAGGER: #{route.options}"
```
dblock: Stray `puts`.

aschuster3: Good catch, I totally forgot to remove that when I cherry-picked over the original commit

dblock:
Squash the commits? @LeFnord I'm cool with this, feel free to merge.

PR #435, opened 2016-05-21
Enhance Readme

stefan-kolb:
Add hint for including parsers with non-Rails applications.


Bugagazavr:
Thanks

dblock:
Excellent.

If you want to be pedantic for the documentation, you would want to add a `for example` otherwise someone might try to `require 'grape-swagger/entitty` when working with `representable` :)

PR #437, opened 2016-05-22
Nitpicking readme :)

stefan-kolb:


README.md
```diff
@@ -39,7 +39,7 @@ gem 'grape-swagger-entity'
 gem 'grape-swagger-representable'
 ```
 
-If you are not using Rails, make sure to load the parser inside your application initialization logic via `require 'grape-swagger/entity'`.
+If you are not using Rails, make sure to load the parser inside your application initialization logic, e.g., via `require 'grape-swagger/entity'`.
```
Bugagazavr: Please add  `require 'grape-swagger/representable'` too.

PR #438, opened 2016-05-22
options[:version] missing, changed to route.version.

texpert:
A small change, because the :version is missing in options, but I've found it in route.options. So better to pass the route to PathString - to have it there for .path and .options

### DIFF
--- a/lib/grape-swagger/doc_methods/path_string.rb
+++ b/lib/grape-swagger/doc_methods/path_string.rb
@@ -13,8 +13,8 @@ module GrapeSwagger
           # set item from path, this could be used for the definitions object
           item = route.path.gsub(%r{/{(.+?)}}, '').split('/').last.singularize.underscore.camelize || 'Item'
 
-          if route.options[:version] && options[:add_version]
-            route.path.sub!('{version}', route.options[:version].to_s)
+          if route.version && options[:add_version]
+            route.path.sub!('{version}', route.version.to_s)
           else
             route.path.sub!('/{version}', '')
           end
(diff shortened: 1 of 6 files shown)


LeFnord:
please use the same pattern for `version` as for the other route methods, see: https://github.com/ruby-grape/grape-swagger/blob/master/lib/grape-swagger/grape/route.rb, thanks

texpert:
Done.

dblock:
This most definitely needs a spec, too.

texpert:
Made a spec, at last :)

texpert:
Seems to be in good shape now.

LeFnord:
👍

lib/grape-swagger/doc_methods/path_string.rb
```diff
@@ -2,7 +2,7 @@ module GrapeSwagger
   module DocMethods
     class PathString
       class << self
-        def build(path, options = {})
+        def build(path, route_version, options = {})
```
dblock: Maybe it'd be cleaner if this received the entire route? So `def build(route, options = {})`?

dblock:
It looks good, still needs a CHANGELOG entry and possibly a README update.

lib/grape-swagger/doc_methods/path_string.rb
texpert: It was my first thought https://github.com/ruby-grape/grape-swagger/pull/438/commits/eca701fad634b797824c9b9051deed1887860136 but the PathString specs would need then some complex route initialization.

dblock: You can make the request an `OpenStruct` in the spec, it will actually look simpler I think. API > spec anyway.

texpert:
Now it is clean, the specs updated, CHANGELOG entry added.

LeFnord:
👍

dblock:
Merged, 👍

PR #440, opened 2016-05-27
support custom DELETE responses with models

hedgesky:
If we create and endpoint with such a description:
```ruby
desc 'Endpoint desc', http_codes: [{ code: 200, message: 'desc', model: ExampleEntity }]
```
the entity won't appear in swagger json because of [this line](https://github.com/ruby-grape/grape-swagger/blob/c1c1040accaa097d2d5656b7663c5b4cf10b61ac/lib/grape-swagger/endpoint.rb#L183). It sees default 204 response (which was built few lines upper) and immediately moves to next entry.
This PR fixes that.

### DIFF
--- a/lib/grape-swagger/endpoint.rb
+++ b/lib/grape-swagger/endpoint.rb
@@ -180,7 +180,7 @@ module Grape
           value[:code] = 204
         end
 
-        next if memo.key?(204)
+        next if value[:code] == 204
         next unless !response_model.start_with?('Swagger_doc') &&
                     ((@definitions[response_model] && value[:code].to_s.start_with?('2')) || value[:model])
 


Bugagazavr:
what you think about skip default codes assigement if http_codes present?

dblock:
This also needs tests and CHANGELOG to be merged.

hedgesky:
@Bugagazavr personally, I like your idea, but it could affect those users who add just error http codes and rely on the library regarding successful cases.

LeFnord:
@hedgesky with this [dsl/desc.rb](https://github.com/ruby-grape/grape/blob/master/lib/grape/dsl/desc.rb#L90-L96) in mind, the actual behavior is correct for such an endpoint:
```ruby
desc 'deleting stuff'
delete ':id' do
  # do it
  body false
end
```
this returns status 204

think, what you want to try to achieve is something like this:
```ruby
desc 'deleting stuff',
  success: ExampleEntity
delete ':id' do
  # do it
  present :something, with: ExampleEntity
end
```

this is also possible, but returns status 200

means: adding a 2xx to the `failures` → `http_codes` array doesn't match the meaning of it, and we should be consistent with grape 😉 

@Bugagazavr think skipping defaults isn't such a good idea, for reason one could see in the example above, and IMO we should document the status code of a response, cause it is always in the response

hedgesky:
@LeFnord Thanks for pointing that `failure` == `http_codes`. Actually, it's written in [`Using an option hash`](https://github.com/ruby-grape/grape-swagger#using-an-options-hash) section of Readme, but I've found out all `http_codes` thing in [`Response documentation`](https://github.com/ruby-grape/grape-swagger#response-documentation) section. And there is no mention about it's related only to errors there.
Of course, it would be better if a developer read whole readme at first, but maybe name like `error_http_codes` would be more meaningful, what do you think?

LeFnord:
@hedgesky … the keys `success/entity` and `failures/http_codes` coming direct from grape,
mean changing of one of them should be done there

feel free to update the README

Issue #429, opened 2016-05-18
Does `base_path` still accept a proc?


Lordnibbler:
Adding grape-swagger to an existing rails project, I want to use a different base path depending on the request host:

```ruby
add_swagger_documentation(
  hide_format: true,
  hide_documentation_path: true,
  base_path: proc { |request| request.host =~ /^api/ ? '/' : '/api' },
)
```

When trying to access the `/docs` route to see my swagger-ui page, I see the following error in Rails:

```sh
NoMethodError (undefined method `host' for nil:NilClass):
  apps/api/lib/myapp/api.rb:43:in `block in <class:API>'
  app/middleware/rack/fix_api_headers.rb:12:in `call'
  app/middleware/rack/geolocation.rb:26:in `call'
  app/middleware/rack/fix_ip_address.rb:43:in `call'
  lib/correlation/rack/timbersaw.rb:14:in `block (2 levels) in call'
  lib/timbersaw.rb:37:in `block in tagged'
  lib/timbersaw/logger.rb:27:in `tagged'
  lib/timbersaw.rb:36:in `tagged'
  lib/correlation/rack/timbersaw.rb:13:in `block in call'
  lib/correlation/rack/timbersaw.rb:12:in `call'
  app/middleware/handle_invalid_percent_encoding.rb:20:in
```

It seems that a request is no longer passed into the `proc` in `base_path`, or am I missing something? Happy to update the documentation with a helpful example if someone can point me in the right direction.

LeFnord:
mmh … use a `lambda` instead and get `request` values from environment
as example, please refer to the specs → [api_swagger_v2_global_configuration_spec.rb](https://github.com/ruby-grape/grape-swagger/blob/master/spec/swagger_v2/api_swagger_v2_global_configuration_spec.rb)

dblock:
Did this use to work? We probably want to support a function that takes `request`, it's a fairly common scenario.

LeFnord:
I agree, but the error here is, that `request`, provided as block param is `nil`,
the question now is, where can the request object can get?

Fryie:
Yes, this used to work: https://github.com/ruby-grape/grape-swagger/blob/6e050fb11081a5352e3521a58fff11205804ea8e/lib/grape-swagger/doc_methods.rb#L304

[...]

PR #441, opened 2016-05-28
accepting proc for options

LeFnord:
this solves https://github.com/ruby-grape/grape-swagger/issues/429

`host` and `base_path` accepts `String`, `proc` or `lambda`


README.md
```diff
@@ -216,6 +220,8 @@ add_swagger_documentation \
    base_path: '/super/api'
 ```
 
+`host` and `base_path` are also accepting a `proc` or `lambda`
```
dblock: Make it into a full sentence, like "The `host` and `base_path` options also accept ..".

lib/grape-swagger/doc_methods/optional_object.rb
```diff
@@ -4,11 +4,23 @@ class OptionalObject
...
+            return evaluate(key, options, request) if options[key].is_a?(Proc)
+            options[key]
           else
-            request
+            request.send(default_values[key])
           end
         end
+
+        def evaluate(key, options, request)
+          options[key].arity == 0 ? options[key].call : options[key].call(request)
```
dblock: I am not in love with this. IMO this should be an API that takes a fixed number of parameters. Maybe we should also make the parameter a hash right away so that we can pass more than just the request in here?

dblock:
I am going to merge this, my comments are more like next steps. We can still change things wrt my arity comment, but I'd be ok with shipping it as is too.

LeFnord:
yeap … good point
to your last comment, also yeap, agreeing, but also thinking issues are not enough for that,
better something like tickets, card, whatever, what is better suitable for developing ideas

README.md
Lordnibbler: In addition to that, I think providing some trivial example would be super helpful in the docs here:

```ruby
add_swagger_documentation(
  base_path: proc { |request| request.host =~ /^api/ ? '/' : '/api' }
)
```

Lordnibbler:
Sure, thanks for addressing this; I left a small comment suggesting we have a usage example for the `base_path` option using a `proc` in the docs so folks dont necessarily have to dig in the tests/source

dblock:
@Lordnibbler do PR that!

Lordnibbler:
Sure

Lordnibbler:
https://github.com/ruby-grape/grape-swagger/pull/442

PR #442, opened 2016-05-29
Update documentation with example of base_path proc

Lordnibbler:
per https://github.com/ruby-grape/grape-swagger/pull/441#issuecomment-222338147

### DIFF
--- a/README.md
+++ b/README.md
@@ -202,7 +202,12 @@ You can pass a hash with optional configuration settings to ```add_swagger_docum
 *not all configuration options supported yet*, but is WIP
 
 
-`host` and `base_path` are also accepting a `proc` to evaluate.
+The `host` and `base_path` options also accept a `proc` or `lambda` to evaluate, which is passed a request object:
+
+```ruby
+add_swagger_documentation \
+  base_path: proc { |foo| foo.host =~ /^example/ ? '/api-example' : '/api' }
+```
 
 <a name="host" />
 #### host:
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -202,7 +202,12 @@ You can pass a hash with optional configuration settings to ```add_swagger_docum
 *not all configuration options supported yet*, but is WIP
 
 
-`host` and `base_path` are also accepting a `proc` to evaluate.
+The `host` and `base_path` options also accept a `proc` or `lambda` to evaluate, which is passed a request object:
+
+```ruby
+add_swagger_documentation \
+  base_path: proc { |foo| foo.host =~ /^example/ ? '/api-example' : '/api' }
```
dblock: `foo` here should be `request`

README.md
```diff
@@ -202,7 +202,12 @@ You can pass a hash with optional configuration settings to ```add_swagger_docum
 *not all configuration options supported yet*, but is WIP
 
 
-`host` and `base_path` are also accepting a `proc` to evaluate.
+The `host` and `base_path` options also accept a `proc` or `lambda` to evaluate, which is passed a request object:
```
dblock: change request to the class name linked to the Rack doc of the type of request

Lordnibbler:
@dblock updated! Thanks for the suggestions. I also updated the test where I copied that `foo` request object from!

dblock:
Thanks, merged.

Issue #443, opened 2016-06-02
override type with documentation parameter


scauglog:
multi types seems not well supported, so I would like to use documentation hash to specify which type the documentation should use. for example
```ruby
params do
  optional :filters, types: [JSON, Hash], desc: "json format", documentation: { type: "json"}
end
   post "private/:module_name/:method_name" do
end
```

So I modified in `lib/grape-swagger/doc_methods/parse_params` the "call" method (line 9 to 13)

before
```ruby
data_type = GrapeSwagger::DocMethods::DataType.call(settings)
additional_documentation = settings[:documentation]
if additional_documentation
  settings = additional_documentation.merge(settings)
end
```

after
``` ruby
additional_documentation = settings.fetch(:documentation, {})
settings.merge!(additional_documentation)
data_type = GrapeSwagger::DocMethods::DataType.call(settings)
```

this work for me. But maybe there is a good reason to merge settings in documentation or to do this merge after calling DataType.

(edit: you shouldn't delete the documentation key after merging)

LeFnord:
@scauglog … yeap, multi types are not supported at the moment
please make a proposal, how it could be documented in with swagger,
and validate it against http://bigstickcarpet.com/swagger-parser/www/index.html

I didn't found a solution yet, but maybe you?

scauglog:
[swagger don't allow multi type](https://github.com/OAI/OpenAPI-Specification/issues/458). But we could allow user to specify the type they prefer when generating the doc or select the first type provided in the Array as the type or both.

LeFnord:
@scauglog … yeap I know, that's why again … make a proposal as PR 😉

scauglog:
I noticed that according to the readme the default value provided in the documentation hash override the grape default `Note that the example value will override the Grape default value.` currently this doesn't work but with my fix it will.

LeFnord:
https://github.com/ruby-grape/grape-swagger/pull/444 solved it

PR #444, opened 2016-06-03
multi type and documentation hash

scauglog:
multi type parameter default to first type
when :type and :default are provided in a documenation hash they override grape :type and :default

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,11 +3,14 @@
 #### Features
 
 * [#448](https://github.com/ruby-grape/grape-swagger/pull/448): Header parameters are now prepended to the parameter list - [@anakinj](https://github.com/anakinj).
+* [#444](https://github.com/ruby-grape/grape-swagger/pull/444): With multi types parameter the first type is use as the documentation type [@scauglog](https://github.com/scauglog)
 
 #### Fixes
 
 * [#450](https://github.com/ruby-grape/grape-swagger/pull/438): Do not add :description to definitions if :description is missing on path - [@texpert](https://github.com/texpert).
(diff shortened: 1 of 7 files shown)


scauglog:
resolve issue #443

CHANGELOG.md
```diff
@@ -2,11 +2,12 @@
 
 #### Features
 
-* Your contribution here.
+* when a parameter as multi types the first type is use as the documentation type
```
LeFnord: please could adept it to the normal changeling format, thanks

LeFnord:
👍 … thanks, good work, please run `bundle exec rake`, so you see the rubocop hints 😉

scauglog:
sorry

LeFnord:
please adept the travis file, this feature is only available for grape >= 0.14.0

scauglog:
I run the multi type test only `if Grape::VERSION >= '0.14'`. Is that good ?

LeFnord:
@scauglog … nice solution ≥ and if a user tries to uses `types` but has only grape 0.13.0, then it raises an error, right?
I'm fine with it …

scauglog:
yes, grape will raise an `unknown validator: types`. The same error that make the test fail for grape 0.13 and 0.12. The problem was that before grape 0.14 you can't use `types` when defining a parameter. I do some research and before grape 0.14 you can use `type` with an array of type but grape will take the first element of the array to check type (and I set the swagger type to the first element of the array). 

so now I test `type: [String, Integer]` for all grape version and `types: [String, Integer]` if grape >= 0.14. All case I can think are tested if you see anything else tell me.

LeFnord:
@scauglog … thanks 👍

PR #447, opened 2016-06-06
Fix for tags for path versioned endpoints

### LINKED ISSUE #446: Path versioned endpoints gets the version as tag instead of namespace.
When using :path versioned endpoints the endpoints gets tagged by version and not by namespace. That makes Swagger UI group all versioned endpoints under the same group.

anakinj:
This should fix #446

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 #### Fixes
 
 * [#450](https://github.com/ruby-grape/grape-swagger/pull/438): Do not add :description to definitions if :description is missing on path - [@texpert](https://github.com/texpert).
-* Your contribution here.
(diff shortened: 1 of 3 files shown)


dblock:
Please check the build and update CHANGELOG.

anakinj:
Updated the Changelog.

The Build failing seemed to have been caused by some kind of connection issue on the runner. Seems to be working now.

LeFnord:
👍 … thanks, for fixing

LeFnord:
@anakinj … can you rebase it …

anakinj:
@LeFnord, this is now rebased

LeFnord:
👍 … thanks

PR #448, opened 2016-06-06
Order of header parameters

anakinj:
Noticed that the order of the parameters had changed at some point, nowadays the header parameters are added to the end of the parameter array.

That results in possible headers are located at the end of the list in Swagger UI.

What do you think of always having the possible header parameters at the beginning of the list?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,11 +2,11 @@
 
 #### Features
 
-* Your contribution here.
+* [#448](https://github.com/ruby-grape/grape-swagger/pull/448): Header parameters are now prepended to the parameter list - [@anakinj](https://github.com/anakinj).
 
 #### Fixes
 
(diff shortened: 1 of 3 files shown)


dblock:
I don't see why not. I would spec it as well.

anakinj:
Added test and changelog entry.

CHANGELOG.md
```diff
@@ -6,7 +6,7 @@
 
 #### Fixes
 
-* Your contribution here.
+* [#448](https://github.com/ruby-grape/grape-swagger/pull/448): Header parameters are now prepended to the parameter list - [@anakinj](https://github.com/anakinj).
```
LeFnord: please move the entry to features, cause it is not a bug 😉

LeFnord:
👍 … after addressing comment

PR #450, opened 2016-06-07
Fixed find_definitions_and_params to not add :description to definiti…

texpert:
…ons if :description is mising on path.

Spec of move_params updated.

### DIFF
--- a/lib/grape-swagger/doc_methods/move_params.rb
+++ b/lib/grape-swagger/doc_methods/move_params.rb
@@ -41,7 +41,7 @@ module GrapeSwagger
 
           move_params_to_new(name, params)
 
-          @definitions[name][:description] = path[:description]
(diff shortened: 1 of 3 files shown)


LeFnord:
👍 … thanks

lib/grape-swagger/doc_methods/move_params.rb
```diff
@@ -41,7 +41,7 @@ def find_definition_and_params(path, verb)
 
           move_params_to_new(name, params)
 
-          @definitions[name][:description] = path[:description]
+          @definitions[name][:description] = path[:description] if path[:description]
```
dblock: Should probably be `if path.key?(:description)`, minor.

texpert: Nope, the :description key with nil value is confusing swagger-ui.

PR #451, opened 2016-06-07
Fix specs for primitive arrays

Bugagazavr:
To add support primitive arrays

https://github.com/ruby-grape/grape-swagger-entity/pull/7


Bugagazavr:
@LeFnord i think this can be merged

LeFnord:
👍

Issue #445, opened 2016-06-03
Why are Hashes not included?


aschuster3:
While parsing through to see why my Hash was not appearing, I found this line: https://github.com/ruby-grape/grape-swagger/blob/master/lib/grape-swagger/endpoint.rb#L234

Basically, when that line says that if a parameter is a Hash, it will not be included.  In a personal branch, I changed the logic to read as follows (emphasis on the added parenthesis):
```ruby
memo[key] = param.last unless (param.last[:type] == 'Hash' || param.last[:type] == 'Array') && !param.last.key?(:documentation)
```
This makes it so that a Hash or an Array is not included if it is not documented instead of not including Hashes ever.  Before I make a PR upstream, I was curious if this was intentional.  I'm using this in conjunction with `grape-entity` and the problem I faced was when I call `ObjectEntity.documentation` as a required param like as follows:
```ruby
params do
  requires :something, using: ObjectEntity.documentation
end
```
With `ObjectEntity` looking like this:
```ruby
class ObjectEntity < Grape::Entity
  expose :object_id, documentation: { type: Integer, documentation: { in: 'body' } }
  expose :object_hash, documentation: { type: Hash, documentation: { in: 'body' } }
end
```
The `object_id` appears just fine, but the `object_hash` does not.

It is worth mentioning that if I state my parameters as follows, it works fine, but doesn't let me have path paramenters appear as I would prefer:
```ruby
desc 'Sample description' do
  params ObjectEntity.documentation
end
```

I'm using the Master branch of `grape-swagger` with all dependencies updated to the latest versions.

dblock:
I would say that if there's no spec for it, it's not intentional.

dblock:
Does #454 address everything? Closing.

aschuster3:
Yep!

PR #453, opened 2016-06-10
Include documented Hashes

aschuster3:
This is opened to address Issue #445 

Because of the way the logic is stated, raw Hashes can never be included.  After digging through, I found that the reason this `unless` exists in the first place is because Hashes and Arrays are replaced with their object references (e.g. `"address"=>{"$ref"=>"#/definitions/someDefinition"}` is preferred over `"address"=>{"type"=>"object"}`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,14 +3,17 @@
 #### Features
 
 * [#448](https://github.com/ruby-grape/grape-swagger/pull/448): Header parameters are now prepended to the parameter list - [@anakinj](https://github.com/anakinj).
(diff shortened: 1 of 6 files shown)


aschuster3:
I just realized that I don't actually test for the functionality that I claim, so I will close this for right now.  I unfortunately won't have access to update the tests for a couple of days, but I will add what's needed.  As a brief summary, this PR specifically affects Hashes and Arrays in JSON bodies, which I failed to cover.  I will reopen a new PR when I get the work straightened out.

LeFnord:
@aschuster3 … 👍

PR #454, opened 2016-06-13
Include documented Hashes

aschuster3:
This is opened to address Issue #445 and is a reopen of PR #453 

Because of the way the logic is stated, raw Hashes can never be included. After digging through, I found that the reason this exists in the first place is because Hashes and Arrays are replaced with their object references when possible (e.g. `"address"=>{"$ref"=>"#/definitions/someDefinition"}` is preferred over `"address"=>{"type"=>"object"}`).

Another addition to this PR that was relevant to the original problem is that `readOnly` parameters are included in a request definition generated by the MoveParams class.  [Open API v2 spec says to exclude readOnly parameters in request objects](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#fixed-fields-13) and seeing as the purpose of MoveParams is to generate a helper request definition, I removed the addition of readOnly parameters.  This specifically addresses redundant path parameters in the JSON block.

Because the problems both of these features are closely related, I felt that the scope of this PR encompasses both changes.


LeFnord:
👍 … awesome :)

dblock:
Excellent work @aschuster3.

PR #455, opened 2016-06-13
Allow passing of Array with brackets (eg. Array[String], Array[Integer])

TySpring:
According to docs in Grape, you can pass a parameter of type array with a nested type, like so:
```ruby
params do
  requires :my_param, type: Array[String], documentation: { param_type: 'body' }
end
```
(See https://github.com/ruby-grape/grape#custom-types-and-coercions section for example).

Before this PR, if you do this, the swagger JSON that is generated looks like this:
```JSON
"my_param": {
    "type": "string",
    "description": "this is my param that is supposed to be an array of strings"
}
```

Now, it looks like this:
```JSON
"my_param": {
    "type": "array",
    "description": "This is my param that is an array of strings",
    "items": {
        "type": "string"
    }
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 #### Fixes
 
+* [#455](https://github.com/ruby-grape/grape-swagger/pull/455): Passing `Array[<primitive>]` should create json indicating `array` type - [@tyspring](https://github.com/tyspring).
 * [#450](https://github.com/ruby-grape/grape-swagger/pull/438): Do not add :description to definitions if :description is missing on path - [@texpert](https://github.com/texpert).
 * [#447](https://github.com/ruby-grape/grape-swagger/pull/447): Version part of the url is now ignored when generating tags for endpoint - [@anakinj](https://github.com/anakinj).
 * [#444](https://github.com/ruby-grape/grape-swagger//pull/444): Default value provided in the documentation hash, override the grape default [@scauglog](https://github.com/scauglog).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 #### Fixes
 
-* [#455](https://github.com/ruby-grape/grape-swagger/pull/455): Passing `Array[<primitive>]` should create json indicating `array` type - [@tyspring](https://github.com/tyspring).
+* [#455](https://github.com/ruby-grape/grape-swagger/pull/455): Setting `type:` option as `Array[Class]` creates `array` type in JSON - [@tyspring](https://github.com/tyspring).
 * [#450](https://github.com/ruby-grape/grape-swagger/pull/438): Do not add :description to definitions if :description is missing on path - [@texpert](https://github.com/texpert).
 * [#447](https://github.com/ruby-grape/grape-swagger/pull/447): Version part of the url is now ignored when generating tags for endpoint - [@anakinj](https://github.com/anakinj).
 * [#444](https://github.com/ruby-grape/grape-swagger//pull/444): Default value provided in the documentation hash, override the grape default [@scauglog](https://github.com/scauglog).
--- a/README.md
+++ b/README.md
@@ -481,6 +481,30 @@ end
 }
 ```
 
+#### Array type
+
+If you want to pass an array of primitive types, you can do this with the following:
+
+```ruby
+params do
+  requires :action_ids, type: Array[Integer]
+end
+post :act do
+  ...
+end
+```
+
+```json
+{
+  "in": "formData",
+  "name": "action_ids",
+  "type": "array",
+  "items": {
+      "type": "integer"
+  },
+  "required": true
+}
+```
 
 #### Multi types
 
--- a/README.md
+++ b/README.md
@@ -483,7 +483,7 @@ end
 
 #### Array type
 
-If you want to pass an array of primitive types, you can do this with the following:
+Array types are also supported.
 
 ```ruby
 params do
--- a/lib/grape-swagger/doc_methods/parse_params.rb
+++ b/lib/grape-swagger/doc_methods/parse_params.rb
@@ -64,7 +64,9 @@ module GrapeSwagger
           if value_type[:is_array]
             if value_type[:documentation].present?
               param_type = value_type[:documentation][:param_type]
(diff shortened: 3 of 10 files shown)


lib/grape-swagger/endpoint.rb
```diff
@@ -235,6 +235,11 @@ def parse_request_params(required)
       end
     end
 
+    def param_type_is_array?(param_type)
+      return false unless param_type
+      param_type == 'Array' || (param_type[0] == '[' && param_type[-1] == ']' && !param_type.include?(','))
```
dblock: This seems a little hacky :) Maybe just a regex? What's the story with the `,`?

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 
 #### Fixes
 
+* [#455](https://github.com/ruby-grape/grape-swagger/pull/455): Passing `Array[<primitive>]` should create json indicating `array` type - [@tyspring](https://github.com/tyspring).
```
dblock: Since this is a changelog, instead of "should create" make it "generates `array` type for Array[Class]" for example.

README.md
```diff
@@ -481,6 +481,30 @@ end
 }
 ```
 
+#### Array type
+
+If you want to pass an array of primitive types, you can do this with the following:
```
dblock: To nitpick, you're not passing anything. Aren't we just saying that array types are supported? Maybe just mention it above?

lib/grape-swagger/doc_methods/parse_params.rb
```diff
@@ -64,7 +64,9 @@ def document_array_param(value_type)
           if value_type[:is_array]
             if value_type[:documentation].present?
               param_type = value_type[:documentation][:param_type]
-              type = GrapeSwagger::DocMethods::DataType.mapping(value_type[:documentation][:type])
+              if (doc_type = value_type[:documentation][:type])
```
dblock: Rubocop should have complained, that `if` should be on the right hand side of the assignment.

dblock:
This is great, thank you. Take a look at my comments, please, mostly minor.

lib/grape-swagger/endpoint.rb
TySpring: @dblock I could do a regex. The comma is because of a tricky situation. In Grape, you can specify type in 2 similar ways:

1. `type: Array[String]`, meaning an array of strings, and
1. `type: [String, Integer]`, meaning a multi-type param of either 'String' or 'Integer'

In the first case, once the options are within the realm of grape-swagger (at least from what I can tell), the `Array` prefix is stripped, and it just looks like `'[String]'`. The only way I could think to disambiguate `'[String]'` (array of type string) from `'[String, Integer]'` was to look for once vs. multiple items inside brackets, hence the ",". Open to other suggestions, but it's all I could think of.

lib/grape-swagger/doc_methods/parse_params.rb
TySpring: Sorry, I don't know how you mean. Instead of above, I could do:

```ruby
if value_type[:documentation][:type]
  doc_type = value_type[:documentation][:type]
  # ...
end
```

Is that better to you?

TySpring:
@dblock thanks for taking a look!

I addressed all the comments, except one (I wasn't sure exactly what you meant on moving the `if` on assignment)

lib/grape-swagger/endpoint.rb
dblock: I would make this more explicit. For example:

```ruby
param_types = param_type.match(/^\[(.*)\]$/)
param_types = param_types[0].split(',') if param_types
# deal with param_types.size == 1
# deal with param_types.size > 1
```

dblock: Also make sure to add tests for the multiple types case.

lib/grape-swagger/doc_methods/parse_params.rb
dblock: There's an assignment inside an `if`, that's not great.

```ruby
doc_type = value_type[:documentation][:type]
type = GrapeSwagger::DocMethods::DataType.mapping(doc_type) if doc_type
```

lib/grape-swagger/endpoint.rb
```diff
@@ -235,6 +235,11 @@ def parse_request_params(required)
       end
     end
 
+    def param_type_is_array?(param_type)
+      return false unless param_type
+      param_type == 'Array' || param_type =~ /\[\w+\]\z/
+    end
```
dblock: You're parsing this twice, perpetuating the mess we have here and adding to it :) Refactor `param_type_is_array` into `parse_param_type` or something like that and return the array key. Nuke the whole `possible_key = param.first.to_s.gsub('[', '[][')` as well as part of that.

lib/grape-swagger/doc_methods/parse_params.rb
TySpring: oh yeah, totally makes sense

dblock:
Just to loop the loop, I feel pretty strongly about that comment above on parsing, lets clean it up in this PR?

lib/grape-swagger/endpoint.rb
LeFnord: @dblock … the whole `('[', '[][')` junk could be removed by avoiding the introducing of it here [endpoint.rb#L224](https://github.com/ruby-grape/grape-swagger/blob/master/lib/grape-swagger/endpoint.rb#L224)

I had to make a decision, should the name of the parameter indicating if it is a Hash, Array or not, this solution following the rails naming convention for form data

questions is, is it really needed … think removing it, could made many things easier

TySpring: so @LeFnord are you suggesting that the output for nested Arrays in formData would look like `address[line1]` as opposed to `address[][line1]` ?

dblock:
I'm merging this, @LeFnord might make more comments and we can iterate from here! Thanks @TySpring for hanging in here.

TySpring:
Alright sounds great, happy to help more if I can. Thanks!

LeFnord:
@TySpring … for the your work
@dblock … it's fine for me, Inwant to start a discussion, som, I don't have to decide it alone 😉

Issue #457, opened 2016-06-14
Issue with camel case endpoints


rayko:
When using camel case on endpoint names, the app throws an error while loading the grape clases and modules:

```
/home/rayko/.rvm/gems/ruby-2.2.5@app/gems/grape-swagger-0.20.3/lib/grape-swagger.rb:90:in `block in combine_namespace_routes': undefined method `reject' for nil:NilClass (NoMethodError)
```

The following method in grape-swagger.rb

```
      def combine_routes(app, doc_klass)
        app.routes.each do |route|
          route_path = route.path
          route_match = route_path.split(/^.*?#{route.prefix.to_s}/).last
          next unless route_match
          route_match = route_match.match('\/([\w|-]*?)[\.\/\(]') || route_match.match('\/([\w|-]*)$')
          next unless route_match
          resource = route_match.captures.first
          next if resource.empty?
          resource.downcase!
          @target_class.combined_routes[resource] ||= []
          next if doc_klass.hide_documentation_path && route.path.match(/#{doc_klass.mount_path}($|\/|\(\.)/)
          @target_class.combined_routes[resource] << route
        end
      end
```

Causes to change the name that is picked up in the next step on the method:

```
      def combine_namespace_routes(namespaces)
        # iterate over each single namespace
        namespaces.each do |name, namespace|
          # get the parent route for the namespace
          parent_route_name = name.match(%r{^/?([^/]*).*$})[1]
          parent_route = @target_class.combined_routes[parent_route_name]
          # fetch all routes that are within the current namespace
          namespace_routes = parent_route.reject do |route|
            !route_path_start_with?(route, name) || !route_instance_variable_equals?(route, name)
          end

          if namespace.options.key?(:swagger) && namespace.options[:swagger][:nested] == false
            # Namespace shall appear as standalone resource, use specified name or use normalized path as name
            identifier =  if namespace.options[:swagger].key?(:name)
                            name.tr(' ', '-')
                          else
                            name.tr('_', '-').gsub(/\//, '_')
                          end
            @target_class.combined_namespace_identifiers[identifier] = name
            @target_class.combined_namespace_routes[identifier] = namespace_routes

            # # get all nested namespaces below the current namespace
            sub_namespaces = standalone_sub_namespaces(name, namespaces)
            sub_routes = sub_routes_from(parent_route, sub_namespaces)
            @target_class.combined_namespace_routes[identifier].push(*sub_routes)
          else
            # default case when not explicitly specified or nested == true
[...]

PR #458, opened 2016-06-15
Fixed camel case issue #457

rayko:
Issue with camel case endpoints [https://github.com/ruby-grape/grape-swagger/issues/457](https://github.com/ruby-grape/grape-swagger/issues/457)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 * [#444](https://github.com/ruby-grape/grape-swagger//pull/444): Default value provided in the documentation hash, override the grape default [@scauglog](https://github.com/scauglog).
 * [#443](https://github.com/ruby-grape/grape-swagger/issues/443): Type provided in the documentation hash, override the grape type [@scauglog](https://github.com/scauglog).
 * [#454](https://github.com/ruby-grape/grape-swagger/pull/454): Include documented Hashes in documentation output - [@aschuster3](https://github.com/aschuster3).
+* [#457](https://github.com/ruby-grape/grape-swagger/issues/457): Issue with camel case endpoints - [@rayko](https://github.com/rayko/).
 * Your contribution here.
 
 ### 0.21.0 (June 1, 2016)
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
 * [#444](https://github.com/ruby-grape/grape-swagger//pull/444): Default value provided in the documentation hash, override the grape default [@scauglog](https://github.com/scauglog).
 * [#443](https://github.com/ruby-grape/grape-swagger/issues/443): Type provided in the documentation hash, override the grape type [@scauglog](https://github.com/scauglog).
 * [#454](https://github.com/ruby-grape/grape-swagger/pull/454): Include documented Hashes in documentation output - [@aschuster3](https://github.com/aschuster3).
+* [#457](https://github.com/ruby-grape/grape-swagger/issues/457): Issue with camel case endpoints - [@rayko](https://github.com/rayko/).
```
dblock: Describe what the issue is (something like "camel case endpoint causes ...").

dblock:
The build has failed, please check it out.

rayko:
Is that better?

There was a failing spec before I put my hands on it: spec/swagger_v2/api_swagger_v2_detail_spec.rb:144

Tried with several ruby versions including the one used in the builder, just in case, but it still fails. I don't know what might be causing it, though it's just a \n missing on a returned value.

lib/grape-swagger.rb
```diff
@@ -65,7 +65,6 @@ def combine_routes(app, doc_klass)
           next unless route_match
           resource = route_match.captures.first
           next if resource.empty?
-          resource.downcase!
```
LeFnord: maybe you can re-use this method [operation_id.rb#L13](https://github.com/ruby-grape/grape-swagger/blob/master/lib/grape-swagger/doc_methods/operation_id.rb#L13)

LeFnord:
@Bugagazavr …removed bug label, cause it isn't one, by using ruby naming conventions 😉

dblock:
Is that spec failing on HEAD?

dblock:
Looks like a newer version of kramdown/rouge is generating slightly different text. Go ahead and just fix the spec (change the `\n` in there) and make the build green and we can merge.

rayko:
Should I include a comment in the changelog about that? Also, according to @LeFnord this is not a bug anymore, so my previous update on the changelog might need to be moved to the feature section instead of leaving it in the fixes section.

dblock:
No need to changelog spec fixes. 

Is this really not a bug? Remove your fix, does the spec pass? If so it just doesn't need a CHANGELOG, but we should keep the spec.

rayko:
It's green now. Rebel spec aside, this addresses the issue I've opened earlier, but the labels were changed, so I was asking about that, I don't handle the labels for this. It is an issue because of that downcase method, but it can also be a new feature to support camel case, depends on who looks at this. Either way, without removing that `downcase`, anyone using camel case will get an error as I've explained in the issue.

dblock:
Excellent. I am merging this, it seems straightforward, cc: @LeFnord

LeFnord:
👍 … and thanks again

rayko:
Thanks for the attention :smile:

Issue #403, opened 2016-04-29
Adding version to path only possible via configuration key 'version'?


cank:
This is for `grape-swagger (0.20.2)`.

Not sure if this is a bug or an undocumented configuration option, but I am unable to get the version to correctly appear in the JSON file under 'paths' unless I do the following:

``` ruby
add_swagger_documentation(
  version: 'v1',
  add_version: true,
)
```

This will produce:

``` json
"paths": {
  "/api/v1/experiments": {
  }
}
```

The configuration instructions in README.md lists `api_version` as a configuration option, but not `version`. This is clearly the intended behavior as shown here:

https://github.com/ruby-grape/grape-swagger/blob/7913a9ecf5856eee2c88992f70cd909b333e92bd/lib/grape-swagger/doc_methods/path_string.rb#L16

and in tests here:

https://github.com/ruby-grape/grape-swagger/blob/master/spec/lib/path_string_spec.rb

Should the documentation be updated to reflect this?

dblock:
Yes definitely. This is probably something that was added recently by @LeFnord. Thanks for contributing!

LeFnord:
thanks for pointing …

LeFnord:
have a look on it … https://github.com/ruby-grape/grape-swagger/pull/405

LeFnord:
corrected behaviour:

* version of the API, also used in a path would be taken from `grape` version
* the documentation version could be set via `doc_version`

this could be confusing, but I think the intention of that is, one have a gem/engine, where the API would be implemented, this could be a semantic version, and should also be documented …

BaneOfSerenity:
the add_swagger_documentation version should be in the upgrade guide as it used to be api_version.

dblock:
@BaneOfSerenity I think you make a good point, would you please contribute?

BaneOfSerenity:
@dblock would be glad to, i should have something by the end of the day. ^.^

PR #462, opened 2016-06-23
Update UPGRADING.md

BaneOfSerenity:
Updated upgrade guide from a change in commit https://github.com/ruby-grape/grape-swagger/commit/99c3d820c2eb4d2731f5cae950775bac78dffca2
Not sure which version this falls under, added under >= 0.21.0

Referenced in Issue #403

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -10,6 +10,18 @@ gem 'grape-swagger'
 gem 'grape-swagger-entity'
 ```
 
+`add_swagger_documentation` has changed from
+``` ruby
+  add_swagger_documentation \
+    api_version: 'v1'
+```
+to
+
+``` ruby
+  add_swagger_documentation \
+    version: 'v1'
+```
+
 ### Upgrading to >= 0.10.2
 
 With grape >= 0.12.0, support for `notes` is replaced by passing a block `detail` option specified. For future compatibility, update your code:


UPGRADING.md
```diff
@@ -10,6 +10,18 @@ gem 'grape-swagger'
...
+`add_swagger_documentation` has changed from
+``` ruby
+  add_swagger_documentation \
+    api_version: 'v1'
+```
+to
+
+``` ruby
+  add_swagger_documentation \
+    version: 'v1'
```
LeFnord: did you tried it? think this isn't possible anymore here, cause the API version should be specified in grape,
here one could only add `doc_version`

BaneOfSerenity: This is what i'm currently using in my project and if that's the case it looks like it forwards it on.

dblock:
@BaneOfSerenity Maybe write a spec for this so we know for sure? Thanks.

UPGRADING.md
LeFnord: Can overwrite the API version sets by grape?
If yes, please write a spec to document it → :+1:

LeFnord:
updated UPGRADE.me, inclusive spec could be found here https://github.com/ruby-grape/grape-swagger/pull/489

PR #463, opened 2016-07-01
Support for hiding parameters from documentation

anakinj:
There is a need to be able to hide some experimental/internal parameters from the documentation but still have the possibility to use declared_params to validate that the parameters have been defined for the endpoint.

This kind of solution would allow the definition of parameters but they would be excluded from the swagger documentation.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -28,7 +28,7 @@ Metrics/AbcSize:
 # Offense count: 3
 # Configuration parameters: CountComments.
 Metrics/ClassLength:
-  Max: 209
+  Max: 216
 
 # Offense count: 10
 Metrics/CyclomaticComplexity:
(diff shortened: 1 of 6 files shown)


anakinj:
The Grape::Endpoint class is really long, and RuboCop is complaining about this(therefore the failing build). Any ideas on how to handle this one?

Bugagazavr:
Just run `rubocop -a` or bump this limit manually,

anakinj:
Ok, so mostly green now. But seems that latest Rack version requires Ruby >=2.2.0.
Do we want to add some kind of Gemfile logic for the tests that selects an older version of Rack if RUBY_VERSION < 2.2.0? Any other suggestions on how to handle this?

LeFnord:
@anakinj … yeap the endpoint class is long and should be shorten by refactoring out functionality, but this should be done in another PR. For the gen versions maybe it could be used a similar mechanism as for grape, where different gemfiles exists and the right one would be loaded depending on ruby or specific gem version.
Thanks for your work 👍

anakinj:
@LeFnord, Should we maybe implement these different Gemfiles as a separate PR, the solution for this particular case was adding some logic to the single Gemfile that is used.

LeFnord:
@anakinj … yeap, agree, it should be done in a different PR

ramses7083:
It is posible to hide require params? In the readme only show this feature for optional param.

PR #464, opened 2016-07-04
fixes body array params

LeFnord:
… corrects the setting of the type for parameters which are given as array, adds missing format

### DIFF
--- a/lib/grape-swagger/doc_methods/move_params.rb
+++ b/lib/grape-swagger/doc_methods/move_params.rb
@@ -83,8 +83,13 @@ module GrapeSwagger
             nested.each { |x| params.delete(x) }
             nested_def_name = GrapeSwagger::DocMethods::OperationId.manipulate(nested_name)
             def_name = "#{name}#{nested_def_name}"
-            properties[nested_name] = { '$ref' => "#/definitions/#{def_name}" }
 
+            if nested.first[:type] && nested.first[:type] == 'array'
+              prepare_nested_types(nested)
+              properties[nested_name] = { type: 'array', items: { '$ref' => "#/definitions/#{def_name}" } }
+            else
+              properties[nested_name] = { '$ref' => "#/definitions/#{def_name}" }
+            end
             prepare_nested_names(nested)
             build_definition(def_name)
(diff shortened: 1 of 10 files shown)


lib/grape-swagger/doc_methods/move_params.rb
```diff
@@ -113,6 +118,17 @@ def build_definition(name, verb = nil)
           name
         end
 
+        def prepare_nested_types(params)
+          params.each do |param|
+            param.tap do |x|
```
dblock: This seems silly, all this does is rename `param` into `x`, wjy don't you just do `do |x|` above or, better leave it as `param`?

lib/grape-swagger/doc_methods/move_params.rb
```diff
@@ -145,10 +161,7 @@ def property_keys
         def movable?(param)
           param[:in] == 'body'
         end
-
-        def deletable?(param)
-          param[:in] == 'body'
-        end
+        alias deletable? movable?
```
dblock: Aren't you supposed to make those symbols?, like `:deletable?` and `:movable?`?

dblock:
My `tap` comment makes sense I think, the rest is fine, merge at will.

lib/grape-swagger/doc_methods/move_params.rb
LeFnord: yeap, would be changed, also in `prepare_nested_names`

lib/grape-swagger/doc_methods/move_params.rb
LeFnord: this is a rubocop recommendation

dblock: Oh that's 2.x syntax, mind blown.

dblock:
Should we be worried that things like `nested_definitions` are modifying `params`?

LeFnord:
only if they did … but here only the params given in body, would be moved/translated into a definition object, this could be seen as a direct JSON specification of the request parameters for POST/PUT/PATCH requests

PR #467, opened 2016-07-05
refoctors building definitions for body params

LeFnord:


spec/lib/move_params_spec.rb
```diff
@@ -7,50 +7,6 @@
...
-      specify { expect { |b| subject.find_post_put(paths, &b) }.to yield_successive_args *expected }
-    end
-  end
-
-  describe 'find_definition_and_params' do
-    specify do
-      subject.instance_variable_set(:@definitions, definitions)
-      subject.find_definition_and_params(found_path[:post], :post)
-      expect(definitions.keys).to include 'InBody'
-      expect(definitions['postRequestInBody'].keys).to_not include :description
```
texpert: Wouldn't this case be lost?

LeFnord: in principal yeap … cause now not the whole set of paths and definitions are the parameters of the method → see here https://github.com/ruby-grape/grape-swagger/pull/467/files#diff-7af39528837333c960830618fb6a52e9R162

maybe I should rename it

LeFnord: sorry … first comment was wrong an dnot related,

no it is always there, see: https://github.com/ruby-grape/grape-swagger/pull/467/files#diff-0a42864b10ca75eb0af2c0fc45139eb0R25

LeFnord: but yeah should adding a spec to proof it, thanks

texpert: Yep, I remember it from here https://github.com/ruby-grape/grape-swagger/pull/450/files :)

LeFnord: yeap … re-added it, thanks 😄

PR #470, opened 2016-07-07
request definitions now inline

LeFnord:


dblock:
Read the code, made sense, merged.

PR #475, opened 2016-07-14
Support collectionFormat documentation for parameters.

dimitertodorov:


LeFnord:
@dimitertodorov … thanks for the work,
but I'm missing specs for complex arrays, such as one could find here: https://github.com/ruby-grape/grape-swagger/blob/master/spec/swagger_v2/api_swagger_v2_param_type_body_nested_spec.rb,
how would you handle this?

and yeap, add entry to the changelog

dimitertodorov:
Thanks @LeFnord I will take a look this week at those specs and resubmit.

jdurand:
@dimitertodorov @LeFnord why was this PR closed?

LeFnord:
@jdurand … think that @dimitertodorov did some re-work on it

jdurand:
OK, but `documentation: { collection_format: 'multi' }` is not yet handled right?
@dimitertodorov this is definitely something I could use, are you still working on this?

Issue #400, opened 2016-04-25
params are not loading when defined both in desc block and separate params block


brahamshakti:
Desc block params are not loading when defined with separate params block, e.g.

    desc "Create a User.", {
            headers: {'XAuthToken': {required: true}},
            params: Entities::UserEntity.documentation
    }
    params do
            optional :extra_parameter, type: String
    end

In above case only `extra_parameter` and `header` field are visible. This worked well previously by showing all desc block params field with separate params block field.

This code works well by displaying `header` field  with optional `extra_parameter` 

    desc "Create a User.", {
            headers: {'XAuthToken': {required: true}},
    }
    params do
            optional :extra_parameter, type: String
    end

but below code shows all params field except header field _(I have removed extra param block)_

    desc "Create a User.", {
            headers: {'XAuthToken': {required: true}},
            params: Entities::UserEntity.documentation
    }

I think this is a bug. What you guys think.

LeFnord:
@brahamshakti … think, these params should internally be added to `route.[route_]params`

LeFnord:
@brahamshakti … I investigated a bit and found no way to do it on grape-swagger side, cause it is a grape option for the desc block, see: [desc.rb](https://github.com/ruby-grape/grape/blob/master/lib/grape/dsl/desc.rb#L12) and we should not overwrite grape behaviour

PR #476, opened 2016-07-15
Fixed move_params_to_new to allow name to be defined as a Symbol

anakinj:
I am trying to use a Entity to describe what is expected inside the request body, while doing this I noticed that when parameters have been defined as body params inside the desc block, grape-swagger causes the app to crash.

### DIFF
--- a/spec/swagger_v2/api_swagger_v2_body_definitions_spec.rb
+++ b/spec/swagger_v2/api_swagger_v2_body_definitions_spec.rb
@@ -6,8 +6,14 @@ describe 'body parameter definitions' do
       class Endpoint < Grape::API
         resource :endpoint do
           desc 'The endpoint' do
-            params body_param: { type: String, desc: 'param', required: false, documentation: { in: 'body' } }
+            headers XAuthToken: {
+              description: 'Valdates your identity',
+              required: true
+            }
+            params body_param: { type: 'String', desc: 'param', documentation: { in: 'body' } },
+                   'body_string_param' => { type: String, desc: 'string_param', documentation: { in: 'body' } }
           end
+
           post do
             { 'declared_params' => declared(params) }
           end
@@ -27,10 +33,11 @@ describe 'body parameter definitions' do
     JSON.parse(last_response.body)
   end
 
-  context 'a definition is generated for the endpoints parameters' do
+  context 'a definition is generated for the endpoints parameters defined within the desc block' do
     specify do
       expect(subject['definitions']['postEndpoint']['properties']).to eql(
-        'body_param' => { 'type' => 'string', 'description' => 'param' }
+        'body_param' => { 'type' => 'string', 'description' => 'param' },
+        'body_string_param' => { 'type' => 'string', 'description' => 'string_param' }
       )
     end
   end
--- a/spec/swagger_v2/api_swagger_v2_body_definitions_spec.rb
+++ b/spec/swagger_v2/api_swagger_v2_body_definitions_spec.rb
@@ -11,7 +11,7 @@ describe 'body parameter definitions' do
               required: true
             }
             params body_param: { type: 'String', desc: 'param', documentation: { in: 'body' } },
-                   'body_string_param' => { type: String, desc: 'string_param', documentation: { in: 'body' } }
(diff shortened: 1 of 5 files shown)


LeFnord:
@anakinj … thanks for it, please add a changeling entry

one question, could you access the parameter in the post block?
see this https://github.com/ruby-grape/grape-swagger/issues/400 issue

anakinj:
Noticed a few other problem cases, fixed in the latest push. Changelog also updated.

spec/swagger_v2/api_swagger_v2_body_definitions_spec.rb
```diff
@@ -6,8 +6,14 @@ module TheBodyApi
       class Endpoint < Grape::API
         resource :endpoint do
           desc 'The endpoint' do
-            params body_param: { type: String, desc: 'param', required: false, documentation: { in: 'body' } }
+            headers XAuthToken: {
+              description: 'Valdates your identity',
+              required: true
+            }
+            params body_param: { type: 'String', desc: 'param', documentation: { in: 'body' } },
+                   'body_string_param' => { type: String, desc: 'string_param', documentation: { in: 'body' } }
```
LeFnord: IMO, this should not be possible, it should be used Ruby >2.x hash style

anakinj: Yeah, and that is not actually not relevant, the problem was that the type was defined as a constant.

anakinj:
about "one question, could you access the parameter in the post block?" not really sure what you mean but this is what i figured out:
The parameter can be accessed via the params hash, but the declared method will not include it for some reason.

LeFnord:
what I mean is, can the parameter be accessed in the grape post block, here:
```ruby
post do
  # can it be accessed?
end
```

anakinj:
This definition:
```
class Endpoint < Grape::API
  format :json
  resource :endpoint do
    desc 'The endpoint' do
      params body_param: { type: 'String', desc: 'param', documentation: { in: 'body' } },
             body_type_as_const_param: { type: String, desc: 'string_param', documentation: { in: 'body' } }
    end
    post do
      { 'params' => params,
        'declared_params' => declared(params) }
    end
  end
  add_swagger_documentation
end
```

passes the following tests:
```
context 'param tests' do
  specify do
    post '/endpoint', '{"body_param": "XYZ"}', 'CONTENT_TYPE' => 'application/json'
    expect(last_response.body).to eql('{"params":{"body_param":"XYZ"},"declared_params":{}}')
  end
  specify do
    post '/endpoint', body_param: "XYZ"
    expect(last_response.body).to eql('{"params":{"body_param":"XYZ"},"declared_params":{}}')
  end
end
```

LeFnord:
cool … it seems, one could add parameters via params key, but they would not be added to the declared params array, thanks for clarification

LeFnord:
👍

anakinj:
What I am trying to get working is to have the Entity describe the content of the body, and so far with no luck. So with the fixes in this PR and the following definition I almost get what i want:

```
class Endpoint < Grape::API
  def self.as_body_param(entity_model)
    documentation = {}
    entity_model.documentation.each do |param_name, param_definition|
      documentation[param_name] = param_definition.merge(documentation: { in: 'body'})
    end
    documentation
  end
  resource :endpoint do
    desc 'The endpoint' do
      params TheBodyApi::Endpoint.as_body_param(::Entities::Something)
    end
    post do
      { 'declared_params' => declared(params) }
    end
  end
  add_swagger_documentation
end
```

It would be nice to be able to somehow tell the endpoint that "The body content should be of type X"

LeFnord:
wouldn't it be easier to add `in: body` to the documentation hash of the exposing of the entity self?
it is enough, to do it on one exposing, see: https://github.com/ruby-grape/grape-swagger/blob/master/spec/lib/move_params_spec.rb#L258

anakinj:
You are right. Thinking to complicated here, but as long as I'm using the current version I have to manipulate the hash a little to workaround the bugs :)

The cleanest solution from swagger point of view would be just to reference a existing model for the body parameter. Now a definition gets generated for every endpoint and model that is defined.

LeFnord:
yeap, fixing the behavior of the params key is on the road map

PR #479, opened 2016-07-25
Fix regex for Array and Multi Type in doc_methods

frodrigo:
Parsing of "[APoint]" should return "APoint"

With current regex it fails:
```
irb(main):001:0> "[APoint]".gsub(/[(\A\[)(\s+)(\]\z)]/, '')
=> "Point"
```


LeFnord:
@frodrigo … thanks, but missing spec and changeling entry

frodrigo:
Updated

LeFnord:
👍

PR #482, opened 2016-07-26
[NOT READY] Check Array nested in Array and Hash

frodrigo:
This commit show fail on nested Arrays

```
              optional :addresses, type: Array do
                optional :country, type: Array[String], documentation: { desc: 'country', in: 'body' }
              end
```


LeFnord:
closed in favour of #683

PR #483, opened 2016-07-26
added support for nicknamed routes

pbendersky:
adds support for nicknamed routes (fixes #481)

PR #485, opened 2016-07-29
Added missing . in changelog.

pbendersky:
Fixed changelog (added missing period) from previous PR (#483)


LeFnord:
😄

dblock:
Btw, we rolled out Danger in Grape, it now warns you when such terrible things as missing periods are in CHANGELOG. See https://github.com/ruby-grape/grape/blob/master/Dangerfile. I am trying to figure out how to make a package that can be reused across all Grape projects.

PR #486, opened 2016-07-30
Use an automated PR linter, danger.systems.

dblock:
I'll iterate on this...


Gemfile
```diff
@@ -15,3 +15,5 @@ if RUBY_VERSION < '2.2.2'
   gem 'rack', '<2.0.0'
   gem 'activesupport', '<5.0.0'
 end
+
+gem 'danger', '~> 2.0'
```
LeFnord: Maybe adding `required: false`, so not to introduce unwanted dependency?

dblock: Good idea!

Bugagazavr:
Looks like danger use redcarpet only,  this not possible with jruby, should be:

```rb
unless RUBY_PLATFORM == 'java'
  gem 'danger', '~> 2.0', require: false
end
```

dblock:
I opened https://github.com/danger/danger/issues/394 for Danger + JRuby, thanks @Bugagazavr.

PR #487, opened 2016-07-31
usese danger 2.1.1

LeFnord:
- cleans up travis


dblock:
I actually think not ignoring failures for ruby-head is a mistake here - it's a moving target. I would re-add rubinius the same way with rbx-2.

LeFnord:
mmh, makes sense, will re-add it … but why did you merged it then?

dblock:
I was trigger-happy, merging on green ;)

Issue #490, opened 2016-08-02
POST and PUT requests add 201 response by default


pezholio:
My app actually returns a 202, as the job is queued and a URL to poll is returned, but the generated Swagger docs return a 201 response, as well as my (correct) 202 response. Is this by design? If so, can I override this?

LeFnord:
hi @pezholio … did I understanding it right, two response objects would be generated in the swagger documentation, one with `201` and a second with `202` as key?

In my understanding of the http status codes, this is the most correct documentation for your described behaviour. `202` for the request was accepted, but performed later and `201` for the final success response, the performing was done – resource created.

please can you provide a pseudo code – perfectly as a spec 😉 - so it can be reproduced and documented, thanks

pezholio:
Hi,

Here's my actual code:

```
module Octopub
  module Datasets
    class Create < Grape::API

      desc 'Creates a dataset for an authenticated user. Returns a Job URL, which you can then poll to check the creation status of a job', http_codes: [
        { code: 202, message: 'OK', model: Octopub::Entities::Job }
      ]
      params do
        requires :dataset, type: Hash do
          requires :name, type: String, desc: 'The name of the dataset'
          optional :description, type: String, desc: 'A short description of the dataset'
          optional :owner, type: String, desc: 'The Github organisation to publish the dataset to'
          optional :publisher_name, type: String, desc: 'The name of the person / organisation publishing the data'
          optional :publisher_url, type: String, desc: 'The website of the person / organisation publishing the data'
          requires :license, type: String, desc: 'The ID of the dataset\'s license', values: ["cc-by", "cc-by-sa", "cc0", "OGL-UK-3.0", "odc-by", "odc-pddl"]
          optional :frequency, type: String, desc: 'How freqently the dataset is updated', values: ['One-off', 'Annual', 'Every working day', 'Daily', 'Monthly', 'Every minute', 'Every quarter', 'Half yearly', 'Weekly']
          optional :schema, type: String, desc: 'The URL of a JSON table schema to validate against your file(s)'
        end
        requires :files, type: Array do
          requires :title, type: String, desc: 'The name of the file'
          optional :description, type: String, desc: 'A short description of the file'
          requires :file, type: File, desc: 'The actual file'
        end
      end
      post :datasets do
        authenticate!
        process_files(params["files"])
        job = CreateDataset.perform_async(params["dataset"], params["files"], current_user.id)

        status 202
        Octopub::Entities::Job.represent(job)
[...]

PR #491, opened 2016-08-04
Add `ignore_defaults` option

pezholio:
As outlined in #490 - it seems that grape-swagger adds a default `201` response for `PUT` and `POST` requests. If a `PUT` or `POST` request doesn't immediately create a resource, the correct response is `202 Accepted`. In this case, we want to ignore the default `201` response, and only specify a `202`. 

There may be other use cases as well, so I've added support for an `ignore_defaults` option which only adds user-specified responses, and ignores the default responses specified in https://github.com/ruby-grape/grape-swagger/blob/master/lib/grape-swagger/doc_methods/status_codes.rb

### DIFF
--- a/lib/grape-swagger/endpoint.rb
+++ b/lib/grape-swagger/endpoint.rb
@@ -190,7 +190,11 @@ module Grape
 
         @definitions[response_model][:description] = description_object(route, markdown)
         # TODO: proof that the definition exist, if model isn't specified
-        memo[value[:code]][:schema] = route.options[:is_array] ? { 'type' => 'array', 'items' => { '$ref' => "#/definitions/#{response_model}" } } : { '$ref' => "#/definitions/#{response_model}" }
+        memo[value[:code]][:schema] = if route.options[:is_array]
+                                        { 'type' => 'array', 'items' => { '$ref' => "#/definitions/#{response_model}" } }
+                                      else
+                                        { '$ref' => "#/definitions/#{response_model}" }
+                                      end
       end
     end
 
--- a/lib/grape-swagger/endpoint.rb
+++ b/lib/grape-swagger/endpoint.rb
@@ -190,11 +190,7 @@ module Grape
 
         @definitions[response_model][:description] = description_object(route, markdown)
         # TODO: proof that the definition exist, if model isn't specified
-        memo[value[:code]][:schema] = if route.options[:is_array]
-                                        { 'type' => 'array', 'items' => { '$ref' => "#/definitions/#{response_model}" } }
-                                      else
-                                        { '$ref' => "#/definitions/#{response_model}" }
-                                      end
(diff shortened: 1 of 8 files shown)


LeFnord:
@pezholio … thanks

please run `bundle exec rake` or `bundle exec rubocop` to see the warning 😉

spec/swagger_v2/api_swagger_v2_ignore_defaults_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+  end
+
+  before do
+    get '/v3/swagger_doc'
+  end
+
+  let(:json) { JSON.parse(last_response.body) }
+
+  it 'only returns one response if ignore_defaults is specified' do
+    expect(json['paths']['/delay_thing']['post']['responses']).to eq('202' => { 'description' => 'OK' })
```
LeFnord: please add also an expectation, which proofs that the default wasn't created:
```
expect(json['paths']['/delay_thing']['post']['responses'].keys).not_to include '201'
```

lib/grape-swagger/endpoint.rb
```diff
@@ -192,14 +190,17 @@ def response_object(route, markdown)
 
         @definitions[response_model][:description] = description_object(route, markdown)
         # TODO: proof that the definition exist, if model isn't specified
-        memo[value[:code]][:schema] = if route.options[:is_array]
-                                        { 'type' => 'array', 'items' => { '$ref' => "#/definitions/#{response_model}" } }
-                                      else
-                                        { '$ref' => "#/definitions/#{response_model}" }
-                                      end
+        memo[value[:code]][:schema] = route.options[:is_array] ? { 'type' => 'array', 'items' => { '$ref' => "#/definitions/#{response_model}" } } : { '$ref' => "#/definitions/#{response_model}" }
```
LeFnord: is this a rubocop recommendation? … it is really a very long line … can you change it back?

pezholio: I couldn't see any other way to cut down the number of lines. A bit of a cheat I know. Happy to change it back, but Rubocop will complain, and I'm not sure there's any way to add the functionality without increasing the number of lines :/

spec/swagger_v2/api_swagger_v2_ignore_defaults_spec.rb
pezholio: No probs 👍

lib/grape-swagger/endpoint.rb
LeFnord: yeap, it is known, change it back and increase the number, please

pezholio: Cool - done! 😄

LeFnord:
👍 … cc/ @Bugagazavr, if you are fine with it, I want to cut the next release

Bugagazavr:
LGTM

PR #492, opened 2016-08-12
Define security requirements on endpoint methods

tomregelink:
Allow security requirements to be defined on endpoint methods. 
As described in the swagger 2.0 spec: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#securityRequirementObject


lib/grape-swagger/endpoint.rb
```diff
@@ -116,6 +117,11 @@ def method_object(route, options, path)
       [route.request_method.downcase.to_sym, method]
     end
 
+    def security_object(route)
+      security = route.options[:security] if route.options.key?(:security)
+      security
```
LeFnord: no need for assigning a variable here

LeFnord:
Thanks @tomregelink … :+1: after addressing comment

CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 #### Features
 
-* Your contribution here.
```
Bugagazavr: Please return this.

dblock: There's one below under Fixes, that's why Danger didn't say anything.

PR #493, opened 2016-08-15
Swagger UI endpoint authorization.

texpert:
This helps to guard Swagger UI endpoints adding 3 new options to swagger documentation (in my case, using WineBouncer gem and Doorkeeper, but it is configurable):         

```ruby
endpoint_auth_wrapper: WineBouncer::OAuth2,
swagger_endpoint_guard: "oauth2 false",
oauth_token: 'doorkeeper_access_token'
```

To display the endpoint only for signed admin users, I am using a lambda (I think a few lambdas like this would go to options as well - I will work it later):

```ruby
  not_admins = lambda { |token=nil| token.nil? || !User.find(token.resource_owner_id).admin? }

  resource :users, desc: "Users' operations" do
    desc 'List users', hidden: not_admins
    oauth2 'admin'
    get do
      render_result User.all
    end
```

Notes:
- Using with Rails, 'rails' and 'wine_bouncer' gems should be required prior to 'grape-swagger' in boot.rb;
- This works with a fresh PR to WineBouncer which is yet unmerged - https://github.com/antek-drzewiecki/wine_bouncer/pull/64.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 #### Features
 
+* [#493](https://github.com/ruby-grape/grape-swagger/pull/493): Swagger UI endpoint authorization. - [@texpert](https://github.com/texpert).
 * [#492](https://github.com/ruby-grape/grape/pull/492): Define security requirements on endpoint methods - [@tomregelink](https://github.com/tomregelink).
 * [#497](https://github.com/ruby-grape/grape-swagger/pull/497): Use ruby-grape-danger in Dangerfile - [@dblock](https://github.com/dblock).
 * Your contribution here.
--- a/README.md
+++ b/README.md
@@ -14,6 +14,7 @@
 * [Model Parsers](#model_parsers)
 * [Configure](#configure)
 * [Routes Configuration](#routes)
+* [Securing the Swagger UI](#oauth)
 * [Markdown](#md_usage)
 * [Response documentation](#response)
 * [Extensions](#extensions)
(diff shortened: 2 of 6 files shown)


LeFnord:
@texpert … nice work, but can you add a spec please

texpert:
@LeFnord , here is the spec, finally :) 

Rubocop's complaining about double negation !!,  it is the only fail (and I think this cop is ambiguos. 
!! looks much better here than some ternary ? to get true or false value) - what do you think?

LeFnord:
@texpert … please rebase and try to use `.present?` to avoid rubocop warning, thanks

LeFnord:
👍 … good work 😄

texpert:
Thanks, @LeFnord!

I am prepairing one last commit to fix some nuances in the README, and after this I will squash.

texpert:
Done!

narze:
For some people wonder on how to use basic auth to protect the endpoint, here's my implementation after some trial & error.

```ruby
class BasicAuthWrapper < Grape::Middleware::Base
  def before
    if !basic_auth(env['HTTP_AUTHORIZATION'])
      context.error!('Unauthorized', 401)
    end
  end

  private

  def basic_auth(auth_header_string)
    return false if !auth_header_string || !auth_header_string.start_with?('Basic ')

    auth_header_string = auth_header_string.gsub(/^Basic /, '')
    decoded_str = Base64.decode64(auth_header_string)
    
	decoded_str == "username:password"
  end
end

# Usage: Add class to `endpoint_auth_wrapper` option
add_swagger_documentation(
  base_path: '/api/partner',
  doc_version: 'v1',
  add_version: true,
  mount_path: '/v1/swagger_doc',
  endpoint_auth_wrapper: BasicAuthWrapper
```

PR #494, opened 2016-08-16
Include headers when body parameters have been defined

anakinj:
Seems that the header parameters is left out from the documentation in some situations, this change will always merge the header parameters to the request parameters.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 #### Fixes
 
+* [#494](https://github.com/ruby-grape/grape-swagger/pull/494): Header parametes are now included in documentation when body parameters have been defined - [@anakinj](https://github.com/anakinj).
 * Your contribution here.
 
 ### 0.23.0 (August 5, 2016)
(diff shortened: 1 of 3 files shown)


Bugagazavr:
@LeFnord looks like workaround, but i prefer to merge all declared params into params all time, without check for `!route.settings[:declared_params].present?`, what do you think?

LeFnord:
@Bugagazavr … with that, we would ignore params which are specified by the `params` key of `desc`, cause which are not declared one, but could be used in other ways

Bugagazavr:
@LeFnord agreed that can be a really problem, how about add a new property, `documentation: { ignore: true }`, to skip unneeded params in documentation?

LeFnord:
@Bugagazavr … this option was added here https://github.com/ruby-grape/grape-swagger/pull/463/files, only the name differs 😉 → `hidden`

anakinj:
So should we just remove all de conditions and always merge the headers?

LeFnord:
@anakinj  … maybe, but not in this PR, it should be done in a different one, cause it needs a wider context

please rebase, so it could be mörged

anakinj:
@LeFnord, this is now rebased and ready to go.

LeFnord:
:+1: … thanks

PR #495, opened 2016-08-18
Spec for checking for formData and body parameters in GET methods

anakinj:
Stumbled upon a little issue when defining parameters for a GET request, at least params of type Array will be presented as "formData" in the swagger documentation.

Swagger UI will exclude the parameter for the GET request because the non-compatible type.

Here is a spec of this case. Any suggestion on how to solve this, or is this already a know issue?

### DIFF
--- a/spec/swagger_v2/api_swagger_v2_params_for_gets_spec.rb
+++ b/spec/swagger_v2/api_swagger_v2_params_for_gets_spec.rb
@@ -1,0 +1,38 @@
+require 'spec_helper'
+
+describe 'parameters should never be of type body or formData for GET methods' do
+  before :all do
+    module TheApi
+      class ParamGetApi < Grape::API
+        resource :resource do
+          desc 'Simple listing of resources'
+          params do
+            optional :string_filter, type: Array(String)
+          end
+          get '/' do
+            []
+          end
+        end
+
+        add_swagger_documentation
+      end
+    end
+  end
+
+  let(:app) { TheApi::ParamGetApi }
+
+  subject do
+    get '/swagger_doc'
+    JSON.parse(last_response.body)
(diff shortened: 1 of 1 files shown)


spec/swagger_v2/api_swagger_v2_params_for_gets_spec.rb
```diff
@@ -0,0 +1,38 @@
+require 'spec_helper'
+
+describe 'parameters should never be of type body or formData for GET methods' do
+  before :all do
+    module TheApi
+      class ParamGetApi < Grape::API
+        resource :resource do
+          desc 'Simple listing of resources'
+          params do
+            optional :string_filter, type: Array(String)
```
LeFnord: shouldn't it be: `Array[String]` instead

anakinj: That might be true. But even with the correct syntax the result is the same.

LeFnord:
if I understand it right, then the spec should fail, but it didn't

anakinj:
aah, thats because it's "pending" at the moment. waiting for a good solution :)

LeFnord:
and btw, it 'fails' only in swagger ui, but the spec allows, such parameter for GET (it would be used the same [operationa object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#operation-object) as for POTS, etc), but yeah, we should ask us if it make sense …

anakinj:
Not really that familiar with the details of the OpenAPI spec, but to me it would make sense to default to something else than formData for GETs. It's of course trivial to define a explicit 'in' parameter for the parameters, so this defaulting would be only a nice feature to have.

PR #501, opened 2016-09-08
adds getting of a specified resource for Rake Tasks

LeFnord:


dblock:
These things should really have tests. Check out https://github.com/digitalplaywright/mongoid-slug/blob/master/spec/tasks/mongoid_slug_rake_spec.rb for an example of how to test Rake tasks.

LeFnord:
Yeap … I know 😉

PR #503, opened 2016-09-19
corrects exposing of inline definitions

LeFnord:
solves:

- https://github.com/ruby-grape/grape-swagger/pull/503
- https://github.com/ruby-grape/grape-swagger-entity/issues/13

### DIFF
--- a/lib/grape-swagger/endpoint.rb
+++ b/lib/grape-swagger/endpoint.rb
@@ -288,7 +288,17 @@ module Grape
     end
 
     def model_name(name)
-      name.respond_to?(:name) ? name.name.demodulize.camelize : name.split('::').last
+      if name.respond_to?(:entity_name)
+        name.entity_name
+      elsif name.to_s.end_with?('Entity', 'Entities')
+        length = 0
+        name.to_s.split('::')[0..-2].reverse.take_while do |x|
+          length += x.length
+          length < 42
+        end.reverse.join
+      else
+        name.name.demodulize.camelize
+      end
     end
 
     def hidden?(route, options)
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -28,7 +28,7 @@ Metrics/AbcSize:
 # Offense count: 3
 # Configuration parameters: CountComments.
 Metrics/ClassLength:
-  Max: 231
+  Max: 240
 
 # Offense count: 10
 Metrics/CyclomaticComplexity:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 
 #### Fixes
 
+* [#503](https://github.com/ruby-grape/grape-swagger/pull/503): Corrects exposing of inline definitions - [@LeFnord](https://github.com/LeFnord).
 * [#494](https://github.com/ruby-grape/grape-swagger/pull/494): Header parametes are now included in documentation when body parameters have been defined - [@anakinj](https://github.com/anakinj).
 * Your contribution here.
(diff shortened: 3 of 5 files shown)


LeFnord:
solves:

- https://github.com/ruby-grape/grape-swagger/pull/503
- https://github.com/ruby-grape/grape-swagger-entity/issues/13

lib/grape-swagger/endpoint.rb
```diff
@@ -288,7 +288,8 @@ def expose_params_from_model(model)
     end
 
     def model_name(name)
-      name.respond_to?(:name) ? name.name.demodulize.camelize : name.split('::').last
+      name = name.to_s.split('::')[0..-2].join('::').constantize if name.to_s.end_with?('Entity', 'Entities')
```
Bugagazavr: Why we can't just drop all '::', and show model name as is?

LeFnord: good point … I did it, cause the name could be very long, see the spec, without it the name would be: `TestDefinitionClass1`

LeFnord: @Bugagazavr … what is your meaning?

Bugagazavr: i think we can trim symbols if maximum defenition length is reached

Bugagazavr: `name.to_s.split('').reverse.take(255).reverse.join`

LeFnord: ok, I'll chang it 😄

Bugagazavr: also we can add `respond_to?(:entity_name)` to customize name manually

LeFnord: where would it it be set?

Bugagazavr: i thin this should be placed in `model_name` method

```rb
if name.respond_to?(:entity_name)
  name.entity_name
else
  name.to_s.gsub('::', '').split('').reverse.take(255).reverse.join
end
```

LeFnord: yeap, but where in the entry/API definition would it be set?

lib/grape-swagger/endpoint.rb
```diff
@@ -288,7 +288,17 @@ def expose_params_from_model(model)
...
+      if name.respond_to?(:entity_name)
+        name.entity_name
+      elsif name.to_s.end_with?('Entity', 'Entities')
+        length = 0
+        name.to_s.split('::')[0..-2].reverse.take_while do |x|
+          length += x.length
+          length < 42
+        end.reverse.join
+      else
+        name.name.demodulize.camelize
```
LeFnord: I let it, cause without it, it would be a breaking change

serggl:
I guess this is a late comment, but why there is 2 different code blocks for parsing model name?

- One that was just updated in this PR: https://github.com/ruby-grape/grape-swagger/blob/master/lib/grape-swagger/endpoint.rb#L290

- and another one here: https://github.com/ruby-grape/grape-swagger/blob/master/lib/grape-swagger/doc_methods/data_type.rb#L43

Second question is what is the magic logic of dropping `Entity` tail in these methods? If I comment this out - specs keep passing...that is confusing

LeFnord:
@serggl … thanks for spotting on it … it seems you have the knowledge to refatore it out,
please can you make a PR, thanks

serggl:
@LeFnord sure. On the way

PR #504, opened 2016-09-20
Added support for set the 'collectionFormat' of arrays

rczjns:
I added support to be able to set the collection format based on swagger specification ( http://swagger.io/specification/#itemsObject ). Now I set the parameter only if it present in the documentation hash, because the default value could be possible braking change.
But I think we could use the default 'csv' format to every arrays. If you agree with that, I will modify my pull request.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 #### Features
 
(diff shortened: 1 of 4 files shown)


LeFnord:
@rczjns … thanks for it, please rebase it
think we should not set a default value, cause it is not required and will be a breaking change

rczjns:
@LeFnord I've rebased it and OK I've kept it without default value.

PR #505, opened 2016-09-21
Combine namespaces with their mount_path to allow APIs with specified mount_paths

KevinLiddle:
I kept getting the following error when I tried to upgrade my grape and grape-swagger libraries:
```
/usr/local/rvm/gems/ruby-2.2.2/gems/grape-swagger-0.20.3/lib/grape-swagger.rb:97:in `block in combine_namespace_routes': undefined method `reject' for nil:NilClass (NoMethodError)
```
which I traced back to how the namespaced routes are collected. The app I'm working on has a bunch of APIs, all mounted with a specified `mount_path`:

```ruby
mount SomeAPI => "/some_api"
```
so this PR should fix the issue of generating proper swagger docs with APIs that have been mounted to a specific route. I think this makes sense, but let me know if I'm mistaken about something here.

### DIFF
--- a/spec/swagger_v2/namespaced_api_spec.rb
+++ b/spec/swagger_v2/namespaced_api_spec.rb
@@ -64,4 +64,56 @@ describe 'namespace' do
       expect(subject['description']).to eql('Description for aspace')
     end
   end
+
+  context 'mounted under a route' do
+    def app
+      namespaced_api = Class.new(Grape::API) do
+        namespace :bspace do
+          get '/', desc: 'Description for aspace'
+        end
+      end
+
+      Class.new(Grape::API) do
+        mount namespaced_api => '/mounted'
+        add_swagger_documentation format: :json
+      end
+    end
+
+    subject do
+      get '/swagger_doc'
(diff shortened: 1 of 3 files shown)


LeFnord:
@KevinLiddle … thanks, this is really cool 👍

spec/swagger_v2/namespaced_api_spec.rb
```diff
@@ -64,4 +64,56 @@ def app
...
+  context 'mounted under a route' do
+    def app
+      namespaced_api = Class.new(Grape::API) do
+        namespace :bspace do
+          get '/', desc: 'Description for aspace'
+        end
+      end
+
+      Class.new(Grape::API) do
+        mount namespaced_api => '/mounted'
```
LeFnord: can you use ruby 2+ Hash style?

KevinLiddle: `namespaced_api` is a variable so I don't think I can

KevinLiddle:
Thanks @LeFnord! Would you happen to have an idea of when this could get released?

PR #510, opened 2016-10-03
Use 'token_owner' instead of 'oauth_token' on Swagger UI endpoint aut…

texpert:
…horization.

Typically the authorization middleware is making available the 'resource_owner' in the endpoint context. so better to use it instead of 'oauth_token' - it will save a request to the DB.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 #### Features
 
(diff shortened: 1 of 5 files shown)


LeFnord:
thanks @texpert

texpert:
My pleasure, @LeFnord ! :)

Issue #507, opened 2016-09-29
Example value feature is broken in swagger ui for body parameters


serggl:
In particular it is broken for parameters which data type is `Grape::Entity`

Here is a maple API and entities definition:
```ruby
class Rest::Sales::Root < Grape::API
  prefix :sales

  version 'v1', using: :path

  module Entities
    class A < Grape::Entity
      expose :a, documentation: {type: 'string'}
      expose :b, documentation: {type: 'number'}
    end
  end

  desc 'Create a new something' do
    success Entities::A
  end

  params do
    optional :data, type: Entities::A
  end

  post 'something' do
    {}
  end

  add_swagger_documentation hide_documentation_path: true, add_version: true, models: [Entities::A]
end
```

And here is how it looks like in swagger:

<img width="967" alt="2016-09-29 16 24 10" src="https://cloud.githubusercontent.com/assets/53304/18956004/7523439e-8662-11e6-8ed5-fc822c846f97.png">

<img width="967" alt="2016-09-29 16 25 00" src="https://cloud.githubusercontent.com/assets/53304/18956009/7e1c4de2-8662-11e6-9554-6eb43b47693c.png">

One the last screenshot you can also notice that data type names are quite different, but they should represent the same type.

Here is a swagger json generated for this API class:
```json
{
	"info": {
		"title": "API title",
		"version": "0.0.1"
	},
	"swagger": "2.0",
	"produces": [
		"application/json"
	],
	"host": "localhost:3000",
	"basePath": "/api",
	"tags": [
		{
			"name": "something",
			"description": "Operations about somethings"
		}
	],
	"paths": {
		"/sales/v1/something": {
			"post": {
				"summary": "Create a new something",
				"description": "Create a new something",
				"produces": [
					"application/json"
				],
				"consumes": [
					"application/json"
				],
				"parameters": [
					{
						"name": "SalesV1Something",
						"in": "body",
						"required": true,
						"schema": {
							"$ref": "#/definitions/postSalesV1Something"
						}
					}
				],
				"responses": {
					"201": {
						"description": "Create a new something",
						"schema": {
							"$ref": "#/definitions/A"
						}
					}
				},
				"tags": [
					"something"
				],
				"operationId": "postSalesV1Something"
			}
		}
	},
	"definitions": {
		"postSalesV1Something": {
			"type": "object",
			"properties": {
				"data": {
					"type": "Rest::Sales::Root::A"
				}
			},
			"description": "Create a new something"
		},
[...]

PR #511, opened 2016-10-03
Fix incorrect data type linking for request params of entity types

serggl:
This is a fix for #507.

Few moments were changed:
1) Refactored to use a single method for parsing entity names (see recent comments to #503)
2) Custom parameter types are now parsed when parameter objects are parsed. This is new.  Previously all custom types were parsed at a later steps. Not sure if we need a `models` option in `add_swagger_documentation` method from now on, since it looks mostly like a workaround
3) changed links to custom types from `"type": "TypeA"` to `"$ref": "#/definitions/TypeA"`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,9 @@ doc
 # bundler
 .bundle
 
+#IDEA temp files
+.idea
+
 # jeweler generated
 pkg
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 #### Fixes
 
 * [#515](https://github.com/ruby-grape/grape-swagger/pull/515): Removes limit on model names - [@LeFnord](https://github.com/LeFnord).
+* [#511](https://github.com/ruby-grape/grape-swagger/pull/511): Fix incorrect data type linking for request params of entity types - [@serggl](https://github.com/serggl).
 * Your contribution here.
 
 ### 0.24.0 (September 23, 2016)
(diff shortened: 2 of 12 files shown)


JanStevens:
One question, why is the length set to 42? Is this somewhere in the json schema spec (or swagger spec)?

serggl:
great question, @JanStevens ! the answer is - no Idea :) this is a part of current code (https://github.com/ruby-grape/grape-swagger/pull/511/files#diff-7af39528837333c960830618fb6a52e9L297) which is not documented

JanStevens:
Yes I know,

The reason why I'm asking is that our current project has different namespaces and internal and public apis. So our entities end up being `AnimalFarm::Internal::Cow` current code strips this all down to `AnimalFarmInternal` (not really useful).

My current fix is to define a `entity_name` class method that just returns the name of the class, that works ofcourse but isn't really handy.

serggl:
Can someone from maintainers bring some light here?

dblock:
We have to ask @LeFnord, this was added in https://github.com/ruby-grape/grape-swagger/commit/054c54fdda1befd24bbe7c038650a1306248480d it looks like.

LeFnord:
Yeap, I did it, cause 42 is the answer of all 😉 …
but seriously, the intention of limiting it was only to have a readable definition name,
which could be very long in some cases, but it seems not long enough …

will create a PR tonight, which remove the limit, ok?

JanStevens:
Maybe also document the `entity_name` option? By default I expect it to be `class.name` and if I want it differently I start adding `entity_name`s to my entities

LeFnord:
one night later … [#515](https://github.com/ruby-grape/grape-swagger/pull/515) removes the limit

@serggl … can you rebase your PR …

serggl:
@LeFnord sure I can. Right after you guys merge #515?

LeFnord:
thanks @serggl … good job 😄

PR #514, opened 2016-10-08
Improve README

lest:
Fix some grammar typos.


LeFnord:
Thank @lest … no need for adding changelog entry

PR #516, opened 2016-10-12
It's 2016 :tada:

Razer6:
Update date in readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1181,4 +1181,4 @@ See [CONTRIBUTING](CONTRIBUTING.md).
 
 ## Copyright and License
 
-Copyright (c) 2012-2014 Tim Vandecasteele and contributors. See [LICENSE.txt](LICENSE.txt) for details.
+Copyright (c) 2012-2016 Tim Vandecasteele and contributors. See [LICENSE.txt](LICENSE.txt) for details.


README.md
```diff
@@ -1181,4 +1181,4 @@ See [CONTRIBUTING](CONTRIBUTING.md).
 
 ## Copyright and License
 
-Copyright (c) 2012-2014 Tim Vandecasteele and contributors. See [LICENSE.txt](LICENSE.txt) for details.
+Copyright (c) 2012-2016 Tim Vandecasteele and contributors. See [LICENSE.txt](LICENSE.txt) for details.
```
dblock: IANAL, but I think we can't just strip copyright from @tim-vandecasteele like that ;)

dblock:
There's a date in LICENSE too.

README.md
LeFnord: I don't mean stripping of Tim … I mean this:
`Copyright (c) 2012-2014 Tim Vandecasteele and contributors; 2015-2016 ruby-grape and contributors`

dblock: I got that, but I don't think it's OK still because it would strip Tim from Copyright for 2015-2016 and I don't think that's right. Let me see if I can have an actual attorney comment.

LeFnord: Tim isn't stripped from it, he is included in contributors (without any contributions), and he is a member of ruby-grape, so it should be enough → make a diff against 1.2 branch and a blame → most of the code has significant changes

dblock: But copyright noice that says it's his copyright till 2015 and not after seems to contradict what @platfrms is saying no? Either way this is just my opinion, do what you need to do :)

LeFnord: think, we should do what @platfrms suggest, its her topic,
to have ruby-grape instead of Tim is only my personal opinion

dblock: I think it should say

```
Copyright (c) 2012-2016 Tim Vandecasteele, ruby-grape, and contributors. See [LICENSE.txt](LICENSE.txt) for details.
```

dblock: Also

platfrms: HOLLA

LeFnord:
would be done here: https://github.com/ruby-grape/grape-swagger/pull/528/files

PR #518, opened 2016-10-12
additions to #515

LeFnord:
… pushed it, right in the moment as it was merged ;)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 
 #### Fixes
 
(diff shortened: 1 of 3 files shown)


LeFnord:
think on this situation: https://github.com/ruby-grape/grape-swagger/blob/master/spec/issues/430_entity_definitions_spec.rb#L45-L59

in the parent entity `entity_name` was defined, all child entities would respond to it, think this isn't expected, with this solution the `entity_name` would only be used where it was defined

dblock:
I feel like it works around inheritance! Seems more "unexpected", but I don't feel that strongly about it.

PR #523, opened 2016-10-20
Allow specifying custom tags at the route level

jordanfbrown:
The current logic to determine a route's set of tags is based solely on the route, which works in most cases, but isn't flexible when you want to override the tags.

The situation that led me here was the following setup:

```
prefix 'locations/:id'

resource :employees do
  desc 'Retrieve employees for a given location'

  get '/' do
    ...
  end
end
```

Which grouped the `/locations/:id/employees` under the `locations` namespace. This was undesirable, because almost all of the endpoints in our API are prefixed with `/locations/:id`, which meant that they were all grouped under the `locations` namespace.

Allowing an additional `tags` attribute on the `desc` method solved the issue for me.

```
desc 'Retrieve employees for a given location', tags: ['employees']
```

Before:

screenshot 2016-10-20 15 47 02

After:

screenshot 2016-10-20 15 47 12

Is there another way to accomplish this that I missed altogether, or does this seem like a reasonable addition?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * [#520](https://github.com/ruby-grape/grape-swagger/pull/520): Response model can have required attributes - [@WojciechKo](https://github.com/WojciechKo).
 * [#510](https://github.com/ruby-grape/grape-swagger/pull/510): Use 'token_owner' instead of 'oauth_token' on Swagger UI endpoint authorization. - [@texpert](https://github.com/texpert).
+* [#523](https://github.com/ruby-grape/grape-swagger/pull/523): Allow specifying custom tags at the route level. - [@jordanfbrown](https://github.com/jordanfbrown).
 * Your contribution here.
 
 #### Fixes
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -4,7 +4,7 @@
 
 * [#520](https://github.com/ruby-grape/grape-swagger/pull/520): Response model can have required attributes - [@WojciechKo](https://github.com/WojciechKo).
 * [#510](https://github.com/ruby-grape/grape-swagger/pull/510): Use 'token_owner' instead of 'oauth_token' on Swagger UI endpoint authorization. - [@texpert](https://github.com/texpert).
-* Your contribution here.
```
Bugagazavr: Please return it after your changelog note

Bugagazavr:
Hello, thank you for your contribution, this is a good feature, can you provide description of this feature in `README.md`?

LeFnord:
thanks @jordanfbrown … did the global tag set also respect your changes?

jordanfbrown:
@Bugagazavr @LeFnord Feedback addressed, thanks!

LeFnord:
👍

PR #527, opened 2016-10-31
accepts string as entity

LeFnord:


dblock:
Looks good, see my rubocop comment otherwise feel free to merge.

PR #531, opened 2016-11-22
UUID data_type format support

migmartri:
Hi!

In the [Swagger spec](http://swagger.io/specification/#dataTypeFormat), uuid is mentioned next to email as a custom format. So I have added support for it.

In any case, have you guys thought about allowing the format customization via `documentation[:format]?`, i.e  `documentation: { type: 'string', format: 'uuid' }`? so hardcoding formats is not needed anymore?

Currently the method `document_type_and_format` seems to to ignore the format for non primitive types.

Thanks!
Miguel

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 #### Features
 
+* [#531](https://github.com/ruby-grape/grape-swagger/pull/531): UUID data_type format support - [@migmartri](https://github.com/migmartri).
 * [#534](https://github.com/ruby-grape/grape-swagger/pull/534): Allows to overwrite defaults status codes - [@LeFnord](https://github.com/LeFnord).
 * Your contribution here.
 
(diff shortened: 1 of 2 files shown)


LeFnord:
thanks @migmartri for contributing, please add a spec
to your idea, think it has two sides, for one it is very fine to have it,
but the downside would be, one can introduce formats which are not supported by OpenAPI/swagger, so how would you handle it?

migmartri:
Thanks @LeFnord 

In my opinion we should proceed the following way:

* Keep the scope of this patch to just add the `uuid` as part of the supported formats since it follows the current approach and `uuid` is explicitly mentioned in the OpenAPI spec.
* Discuss via [another issue](
https://github.com/ruby-grape/grape-swagger/issues/536) the format override feature.

Please note that I have not added tests for my new type since I could not find any existing spec for the method `#primitives` or `PRIMITIVE_MAPPINGS` and I have only extended a constant value. If you still want me to work on extending the current `spec/lib/data_type_spec.rb ` let me know.

Thanks,
Miguel

LeFnord:
answered on referenced issue

LeFnord:
@migmartri yeap, sounds good to me, so I will mörge this one, and the more general approach would be a new one

migmartri:
@LeFnord thank you. I appreciate it :)

Issue #539, opened 2016-11-28
expose Array of objects on POST also in body


LeFnord:
**related**: https://github.com/ruby-grape/grape-entity/issues/252

PR #540, opened 2016-11-28
issue 539: corrects exposing of array in post body

LeFnord:
**solves**: https://github.com/ruby-grape/grape-swagger/issues/539

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 #### Fixes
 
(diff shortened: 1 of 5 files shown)


lumean:
@LeFnord many thanks for the quick fix, I checked with the latest version, now the array works. Also the param_type body seems to work partially, i.e. the parameter is no longer displayed as formData, but the problem is now swagger doesn't display any parameters at all:

image

here is the file i used to test:   
[api_entity_demo.txt](https://github.com/ruby-grape/grape-swagger/files/619029/api_entity_demo.txt)

shall I open a new issue for this?

LeFnord:
no think not, actual there is one similar issue open

Issue #267, opened 2015-06-28
Nested dynamic napespace problem (crash)


igneus:
Let's have a `config.ru` defining a little API doing something with dates:

``` ruby
require 'grape'
require 'grape-swagger'

class API < Grape::API
  segment '/:year' do
    get do
      "Year #{params[:year]}"
    end

    segment '/:month' do
      get do
        "Month #{params[:year]}/#{params[:month]}"
      end

      get '/:day' do
        "Day #{params[:year]}-#{params[:month]}-#{params[:day]}"
      end
    end
  end

  add_swagger_documentation
end

run API
```

On startup `add_swagger_documentation` crashes:

```
/home/igneus/.rvm/gems/ruby-2.1.1/gems/grape-swagger-0.10.1/lib/grape-swagger.rb:70:in `block in combine_namespace_routes': undefined method `collect' for nil:NilClass (NoMethodError)
        from /home/igneus/.rvm/gems/ruby-2.1.1/gems/grape-swagger-0.10.1/lib/grape-swagger.rb:65:in `each'
        from /home/igneus/.rvm/gems/ruby-2.1.1/gems/grape-swagger-0.10.1/lib/grape-swagger.rb:65:in `combine_namespace_routes'
        from /home/igneus/.rvm/gems/ruby-2.1.1/gems/grape-swagger-0.10.1/lib/grape-swagger.rb:39:in `add_swagger_documentation'
        from /home/igneus/tmp/config.ru:21:in `<class:API>'
        from /home/igneus/tmp/config.ru:4:in `block in <main>'
        from /home/igneus/.rvm/gems/ruby-2.1.1/gems/rack-1.6.1/lib/rack/builder.rb:55:in `instance_eval'
        from /home/igneus/.rvm/gems/ruby-2.1.1/gems/rack-1.6.1/lib/rack/builder.rb:55:in `initialize'
        from /home/igneus/tmp/config.ru:in `new'
        from /home/igneus/tmp/config.ru:in `<main>'
        from /home/igneus/.rvm/gems/ruby-2.1.1/gems/rack-1.6.1/lib/rack/builder.rb:49:in `eval'
        from /home/igneus/.rvm/gems/ruby-2.1.1/gems/rack-1.6.1/lib/rack/builder.rb:49:in `new_from_string'
        from /home/igneus/.rvm/gems/ruby-2.1.1/gems/rack-1.6.1/lib/rack/builder.rb:40:in `parse_file'
[...]

PR #543, opened 2016-11-29
Fix mounting APIs in route_param namespaces

milgner:
This hopefully closes [#267](https://github.com/ruby-grape/grape-swagger/issues/267),
which describes a scenario like:

    route_param :vendor do
      mount Endpoints::Events
    end

The fix is very barebone but resolved the issue for me. Please note that I don't have any idea what I'm doing :wink: Both Grape itself as well as the Swagger integration are very complex and the only thing I did was to look at exceptions and fix the incoming data in a way that would make them go away and finally generate some documentation :laughing:

Everything looks like it's supposed to look but there may be other scenarios that are failing now. So feel free to suggest improvements or add amendments as necessary.

### DIFF
--- a/lib/grape-swagger.rb
+++ b/lib/grape-swagger.rb
@@ -96,16 +96,25 @@ module Grape
         end
       end
 
+      def determine_namespaced_routes(name, parent_route)
+        if parent_route.nil?
+          @target_class.combined_routes.values.flatten
+        else
+          parent_route.reject do |route|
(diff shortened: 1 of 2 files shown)


milgner:
@LeFnord alright, no problem, I'll add specs and a Changelog entry (probably tomorrow evening).

LeFnord:
@milgner … some progress on it?

LeFnord:
hi @milgner … again, any progress on it, would be nice to mörge it

PR #551, opened 2016-12-13
Remove dots from operation id

frodrigo:


LeFnord:
thanks @frodrigo … please could you add a spec to have the use case documented

frodrigo:
@LeFnord can you point me where I need to add a test in the spec ?

LeFnord:
@frodrigo sure … you should a test here https://github.com/ruby-grape/grape-swagger/blob/master/spec/lib/operation_id_spec.rb as unit tests, and maybe a request spec in this dir https://github.com/ruby-grape/grape-swagger/tree/master/spec/swagger_v2 this one is more for documentation purpose, how it is possible to have a dot there - I of myself have no idea 😉

frodrigo:
I add a test and amend my commit.

spec/swagger_v2/operation_id_api_spec.rb
```diff
@@ -0,0 +1,25 @@
+require 'spec_helper'
+
+describe 'an operation id api' do
+  def app
+    Class.new(Grape::API) do
+      version '0.1'
```
LeFnord: ah, here it can come from 😉

PR #562, opened 2017-01-04
The guard method should not just pass the second item as a string.

tim-vandecasteele:
If you want to use the guard method as instructed in the readme, you will actually pass the string “false” to the oauth2 method, iso actually passing `false`. As this is a user added string, we should just instance_eval it.

This is clearly made to be used with wine_bouncer, but in the current implementation I so no way possible to actually pass false (so make wine_bouncer see this as public iso the scope :false)

This does change the current behaviour though!


LeFnord:
thanks @tim-vandecasteele … 
@texpert please can you have a look on it, thanks

texpert:
@LeFnord, @tim-vandecasteele, I am currently abroad in vacation, will return home January, 10 and will take a look after this.

LeFnord:
it seems to be ok

Issue #572, opened 2017-01-22
is_array is applied to all possible responses. 


toshe:
There is a problem when defining the responses as an array in the newest version of grape-swagger.
If you define an endpoint as an array then all the possible responses (success & failure) will be generated as references to an array in the swagger doc. However, a single error entity is not an array. 
Currently there is no way to know if all the possible responses are an array or just one.

```ruby
 desc 'Get many belly rubs!',
               {
                   is_array: true,
                   nickname: 'getBellyRubs',
                   success: Entities::BellyRubsEntity, 
                   failure: [[401, 'BadKittenError', SomeErrorEntity],
                             [404, 'NoTreatsError', SomeErrorEntity],
                             [429, 'TooManyScratchesError', SomeErrorEntity]
                   ]
          }
```

here's the generated code: 

```json
"produces":[
   "application/json"
],
"responses":{
   "200":{
      "description":"Get many belly rubs!",
      "schema":{
         "type":"array",
         "items":{
            "$ref":"#/definitions/bellyrub"
         }
      }
   },
   "401":{
      "description":"BadKittenError",
      "schema":{
         "type":"array",
         "items":{
            "$ref":"#/definitions/error"
         }
      }
   },
   "404":{
      "description":"NoTreatsError",
      "schema":{
         "type":"array",
         "items":{
            "$ref":"#/definitions/error"
         }
      }
   },
   "429":{
      "description":"TooManyScratchesError",
      "schema":{
         "type":"array",
         "items":{
            "$ref":"#/definitions/error"
         }
      }
   }
}
```

The 200 response is correctly rendered as an array but the 401, 404 and 429 Error responses that use a different entity are also rendered as array where in fact they should be a single reference.

LeFnord:
@toshe … yeap, I agree with the key `is_array` should only be used for the success entity

toshe:
@LeFnord danke!

PR #574, opened 2017-01-25
issue 572: is_array should only be applied to success

LeFnord:
solves https://github.com/ruby-grape/grape-swagger/issues/572

### DIFF
--- a/spec/issues/572_array_post_body_spec.rb
+++ b/spec/issues/572_array_post_body_spec.rb
@@ -1,0 +1,49 @@
+require 'spec_helper'
+
+describe '#572 is_array is applied to all possible responses' do
+  include_context "#{MODEL_PARSER} swagger example"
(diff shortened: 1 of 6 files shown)


spec/issues/572_array_post_body_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+  let(:app) do
+    Class.new(Grape::API) do
+      namespace :issue_572 do
+        desc 'get issue',
+             is_array: true,
+             success: Entities::UseResponse,
+             failure: [
+               [401, 'BadKittenError', Entities::ApiError],
+               [404, 'NoTreatsError', Entities::ApiError],
+               [429, 'TooManyScratchesError', Entities::ApiError]
```
toshe: :smile_cat:

Issue #584, opened 2017-02-27
path_string.rb mutates the route path


axfcampos:
Hi there,

The `build` method in in the [path_string.rb](https://github.com/ruby-grape/grape-swagger/blob/master/lib/grape-swagger/doc_methods/path_string.rb) class mutates the route path forever when generating swagger documentation.

For example:
A path that has a dynamic field:
`/foo/:dynamic/bar`

Is forever changed into:
`/foo/{dynamic}/bar`

This sort of in place mutations can cause weird side effects for the `grape` gem and its dependencies. 

For example, this mutation completely breaks the `grape-route-helpers` gem has it becomes unable to identity [dynamic fields in the path](https://github.com/reprah/grape-route-helpers/blob/master/lib/grape-route-helpers/decorated_route.rb#L122). Using the generated methods to call specific paths no longer works because of it.

I have code that depends on `grape-route-helpers` to generate the paths on the links in a json HAL response. After calling the swagger documentation endpoint, all endpoints start failing on response serialisation.

I tried [changing the code](https://github.com/ruby-grape/grape-swagger/blob/master/lib/grape-swagger/doc_methods/path_string.rb#L6) to copy the path with something like `path = route.path.dup` and it prevents this problem, but it breaks several parts of the code in this gem. Around 68 specs started failing with this change.

I'm a bit lost on finding a fix for this problem, as of now, this is stopping us from using grape-swagger on our live systems. We can only get a static output using a development version.

LeFnord:
@axfcampos … please can you provide a spec to re-produce it, thanks

axfcampos:
Hi @LeFnord 😄 

```ruby
      specify 'The original route path is not mutated' do
        route = Struct.new(:version, :path).new
        route.path = '/foo/:dynamic/bar'
        subject.build(route, add_version: true)
        expect(route.path).to eq '/foo/:dynamic/bar'
      end
```

This was the spec I made to show the route is mutated, [here is the branch](https://github.com/axfcampos/grape-swagger/tree/fix-build-path-mutating-routes).

This specific mutation is this one: https://github.com/ruby-grape/grape-swagger/blob/master/lib/grape-swagger/doc_methods/path_string.rb#L12

LeFnord:
@axfcampos … please can you give https://github.com/ruby-grape/grape-swagger/pull/585 a try

axfcampos:
Thank you so much @LeFnord, just tested and it works now 🙌  🙇

[...]

PR #585, opened 2017-02-27
issue #584: do not mutate route.path

LeFnord:


lib/grape-swagger/doc_methods/move_params.rb
```diff
@@ -8,15 +8,15 @@ def can_be_moved?(params, http_verb)
           move_methods.include?(http_verb) && includes_body_param?(params)
         end
 
-        def to_definition(params, route, definitions)
+        def to_definition(path, params, route, definitions)
```
timlapluie: Wouldn't it be better to append the new param `path` instead of prepending it? This is a public method so the change could break other code.

spec/lib/path_string_spec.rb
```diff
@@ -6,7 +6,14 @@
   specify { expect(subject).to eql GrapeSwagger::DocMethods::PathString }
   specify { expect(subject).to respond_to :build }
 
-  describe 'operation_id_object' do
+  describe 'path_string_object' do
+    specify 'The original route path is not mutated' do
+      route = Struct.new(:version, :path).new
+      route.path = '/foo/:dynamic/bar'
+      subject.build(route, add_version: true)
```
florianpilz: I would like to see an expectation here as well to see that the `path` was modified, but `route.path` was not.

lib/grape-swagger/endpoint.rb
```diff
@@ -227,11 +227,11 @@ def apply_success_codes(route)
       [default_code]
     end
 
-    def tag_object(route)
+    def tag_object(route, path)
       version = GrapeSwagger::DocMethods::Version.get(route)
       version = [version] unless version.is_a?(Array)
       Array(
-        route.path.split('{')[0].split('/').reject(&:empty?).delete_if do |i|
+        path.split('{')[0].split('/').reject(&:empty?).delete_if do |i|
```
thomet: I know you don't changed that, but please can we extract this split and reject into a separate method which explains, what it did? It's really hard to understand and a good method name would help with that.

wteuber: This array initialisation is quite complex and worth abstracting away from Application logic.

PR #586, opened 2017-02-27
Parameters delimited by dash cause exception

risa:
In our project we are using api calls to return a range from array, like this:

`get '/range/:range_start-:range_end'`

Currently this causes 
`NoMethodError: undefined method `delete!' for nil:NilClass
./lib/grape-swagger/doc_methods/operation_id.rb:22:in `manipulate'`

See the spec below for demostration. 

The problem is caused by difference between regexp `/\-(\w)/` and include?('-'). The path gets translated to operation: "Range_parameterRange{range_start}-{range_end}". The include?('-') matches the string but the regexp does not. And per specs the gsub! returns [nil if no substitutions were made](http://ruby-doc.org/core-2.4.0/String.html#method-i-gsub-21) (not the least surprise principle for me). 

We are using only dash, possibly underscore can be offending as well, but as a first attempt I suggest only dash. 

Suggestions and feedback welcome. Thanks.

### DIFF
--- a/spec/issues/range_parameter_delimited_by_dash_spec.rb
+++ b/spec/issues/range_parameter_delimited_by_dash_spec.rb
@@ -1,0 +1,25 @@
+require 'spec_helper'
+
+describe '#XXX nested entity given as string' do
+  let(:app) do
+    Class.new(Grape::API) do
+      namespace :range_parameter do
+
+        desc 'Get a array with range'
+        get '/range/:range_start-:range_end' do
+          present []
+        end
+      end
+
+      add_swagger_documentation format: :json
+    end
+  end
+
+  subject do
+    get '/swagger_doc'
+    result = JSON.parse(last_response.body)['paths']
+    result
+  end
+
+  specify { expect(subject.keys).to include '/range_parameter/range/{range_start}-{range_end}' }
+end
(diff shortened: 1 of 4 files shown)


spec/issues/range_parameter_delimited_by_dash_spec.rb
```diff
@@ -0,0 +1,24 @@
+require 'spec_helper'
+
+describe '#XXX nested entity given as string' do
```
LeFnord: please can you adapt the description, e.g. https://github.com/ruby-grape/grape-swagger/blob/master/spec/issues/582_file_response_spec.rb#L3 and also rename the spec file itself,
starting with issue number, thanks

spec/issues/range_parameter_delimited_by_dash_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+    end
+  end
+
+  subject do
+    get '/swagger_doc'
+    result = JSON.parse(last_response.body)['paths']
+    result
+  end
+
+  specify { expect(subject.keys).to include '/range_parameter/range/{range_start}-{range_end}' }
```
LeFnord: can you show the `operationId`, how this had changed, thanks

spec/issues/range_parameter_delimited_by_dash_spec.rb
risa: Sorry for the naming, my first PR ever ... should I open an issue for this, to get the number?

LeFnord:
[please add changelog entry, as the bot suggest

LeFnord:
thanks again

risa:
Thanks as well!

PR #588, opened 2017-02-28
Allow extension keys in Info Object

mattyr:
per http://swagger.io/specification/#infoObject

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 #### Features
 
 * [#583](https://github.com/ruby-grape/grape-swagger/pull/583): Issue #582: document file response - [@LeFnord](https://github.com/LeFnord).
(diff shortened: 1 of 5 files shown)


mattyr:
made the requested changes -- also made the extension format conform better with existing extensions (```x: {foo: 'bar'}``` rather than ```"x-foo": 'bar'```).

quick question re: rubocop: i was getting a ton of errors so i thought that was ignored.  looks like 99% are "Missing frozen string literal comment".  I disabled that using .rubocop.yml file and just addressed those issues specific to these changes.  Would you like a PR that either:

- disables the "Missing frozen string literal comment" test universally, or
- adds the comment to all needed files (not sure if this will introduce new spec failures)

LeFnord:
@mattyr … its a bit curious that this cop pops up, cause it is running some weeks, but prefer not to mix up your PR with a maintainance PR, and yeah here is it green

PR #589, opened 2017-02-28
allow overriding tag definitions

mattyr:
So you're not stuck with "Operations about xxxxxx"

### DIFF
--- a/spec/swagger_v2/default_api_spec.rb
+++ b/spec/swagger_v2/default_api_spec.rb
@@ -139,4 +139,28 @@ describe 'Default API' do
       expect(subject['x-logo']).to eql('http://logo.com/img.png')
     end
   end
+
+  context 'with tags' do
+    def app
+      Class.new(Grape::API) do
+        format :json
+        desc 'This gets something.'
+        get '/something' do
+          { bla: 'something' }
+        end
+        add_swagger_documentation tags: [
+          { name: 'something', description: 'customized description' }
+        ]
+      end
+    end
+
+    subject do
+      get '/swagger_doc'
+      JSON.parse(last_response.body)
+    end
+
+    it 'documents the customized tag' do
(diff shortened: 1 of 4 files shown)


LeFnord:
thanks @mattyr … but please have in mind, that the generation of the tags in grape-swagger are strongly related to the paths and an overriding of it is actual implemented, see: [overriding-the-tags](https://github.com/ruby-grape/grape-swagger#overriding-the-tags)

mattyr:
I did see that, however what I was looking to do was decouple that logic (if wanted).  As it stands, it appears that the tag description is still fixed as "Operations about xxxxx", even if you reassign arbitrary tags to a path (correct me if i'm wrong!).  The goal here is to still allow the existing "overriding-the-tags" (just re-assigning the tags), but allow the actual tag definitions (primarily the description) to be customized.

spec/swagger_v2/default_api_spec.rb
```diff
@@ -139,4 +139,28 @@ def app
...
+    def app
+      Class.new(Grape::API) do
+        format :json
+        desc 'This gets something.'
+        get '/something' do
+          { bla: 'something' }
+        end
+        add_swagger_documentation tags: [
+          { name: 'something', description: 'customized description' }
+        ]
```
LeFnord: please can you add a second tag, so that it will be clear, that an existing tag can be overwritten and non-existing will be added, thanks

mattyr: you're right, pushing a revised spec

LeFnord:
@mattyr … please can you force push, to start travis, thanks

mattyr:
done -- @LeFnord out of curiosity, do you not have access to the travis ci to just restart the builds there?

LeFnord:
Hi @mattyr yeap I have access, but there was no build to restart  … say only s3 😉

mattyr:
haha gotcha.  good ol' aws.

Issue #591, opened 2017-03-01
DELETE method regression


mattyr:
Looks like #583 introduced a regression, where if you have a 200 response code with a model (entity) on a delete method, you get:

```sh
NoMethodError - undefined method `[]=' for nil:NilClass:
   () Users/matt/.rvm/gems/ruby-2.2.6@gorecess/bundler/gems/grape-swagger-e1cd8ae702f6/lib/grape-swagger/endpoint.rb:211:in `block in response_object'
   () Users/matt/.rvm/gems/ruby-2.2.6@gorecess/bundler/gems/grape-swagger-e1cd8ae702f6/lib/grape-swagger/endpoint.rb:193:in `response_object'
   () Users/matt/.rvm/gems/ruby-2.2.6@gorecess/bundler/gems/grape-swagger-e1cd8ae702f6/lib/grape-swagger/endpoint.rb:117:in `method_object'
   () Users/matt/.rvm/gems/ruby-2.2.6@gorecess/bundler/gems/grape-swagger-e1cd8ae702f6/lib/grape-swagger/endpoint.rb:97:in `block in path_item'
   () Users/matt/.rvm/gems/ruby-2.2.6@gorecess/bundler/gems/grape-swagger-e1cd8ae702f6/lib/grape-swagger/endpoint.rb:91:in `path_item'
   () Users/matt/.rvm/gems/ruby-2.2.6@gorecess/bundler/gems/grape-swagger-e1cd8ae702f6/lib/grape-swagger/endpoint.rb:76:in `block in path_and_definition_objects'
   () Users/matt/.rvm/gems/ruby-2.2.6@gorecess/bundler/gems/grape-swagger-e1cd8ae702f6/lib/grape-swagger/endpoint.rb:74:in `path_and_definition_objects'
   () Users/matt/.rvm/gems/ruby-2.2.6@gorecess/bundler/gems/grape-swagger-e1cd8ae702f6/lib/grape-swagger/doc_methods.rb:54:in `block in setup'
   () Users/matt/.rvm/gems/ruby-2.2.6@gorecess/bundler/gems/grape-swagger-e1cd8ae702f6/lib/grape-swagger/doc_methods.rb:68:in `block in setup'
  grape (0.17.0) lib/grape/endpoint.rb:59:in `block (2 levels) in generate_api_method'
```

will submit a pr with reproduction/fix in a moment

LeFnord:
[...]

PR #592, opened 2017-03-01
Fixes issue #591: delete method regression

mattyr:


spec/issues/591_delete_method_regression_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+    JSON.parse(last_response.body)
+  end
+
+  specify do
+    expect(subject['paths'].map { |_, p| p['delete']['responses'] }).to eql(
+      [
+        { '200' => { 'description' => 'deleted', 'schema' => { '$ref' => '#/definitions/Foo' } } },
+        { '200' => { 'description' => 'deleted', 'schema' => { '$ref' => '#/definitions/Foo' } } },
+        { '203' => { 'description' => 'deleted' }, '200' => { 'schema' => { '$ref' => '#/definitions/Foo' } } }
+      ]
```
LeFnord: a bit curious, you have 3 endpoints → routes, but 4 success responses

spec/issues/591_delete_method_regression_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+        desc 'delete with model, code 200',
+             is_array: false,
+             success: { code: 200, model: Foo, message: 'deleted' }
+        delete '/deleteme200' do
+          present({ something: 'bar' }, with: Foo)
+        end
+
+        desc 'delete with model, code 204',
+             is_array: false,
+             success: { code: 204, model: Foo, message: 'deleted' }
```
LeFnord: the default status code for delete is 204, with an empty body, but here you are suggesting to respond an entity `Foo` for an empty body 204

mattyr: Correct.  The existing code (prior to this change) modified the 204 to be a 200 if a response body (model) was given.  I maintained that behavior here (ie, I'm not introducing something new here).

spec/issues/591_delete_method_regression_spec.rb
mattyr: it's another attempt to maintain the existing behavior.  technically it's a bad definition, but i didn't want to introduce a brand new regression.  in this scenario the "correct" response definition is appended.  if it's safe to modify this behavior, i can do so.

lib/grape-swagger/endpoint.rb
```diff
@@ -198,7 +198,7 @@ def response_object(route)
         response_model = expose_params_from_model(value[:model]) if value[:model]
 
         if route.request_method == 'DELETE' && !value[:model].nil?
```
LeFnord: habe this change (https://github.com/ruby-grape/grape-swagger/pull/583/files#diff-7af39528837333c960830618fb6a52e9L192) must be reverted

mattyr: Sorry, I'm not sure what you mean, do you want me to revert that change?  That's part of the change that introduced the regression.  But I think that altered behavior intentionally, correct?  If not, then I can revert that change, and I think my new unit tests will still pass; however that means that the new behavior introduced by that line will be reverted back to what it was... is that acceptable?  (sorry if i'm misunderstanding you!)

mattyr:
rebased

lib/grape-swagger/endpoint.rb
LeFnord: yeap, think for the moment – support grape < 0.19.0 – it is the best, the default of 204 for DELETE was introduced in 0.19.0

mattyr: sorry, i'm still confused about how you'd like to address the problem.  i'm willing to make alterations to this pull request.  do you want to:

a) Assert compatibility via a dependency lock on < 0.19.0, and make the changes specific to that version
b) Revert the change that I mention above and restore compatibility with all versions
c) Something else?

Sorry for being confused...

LeFnord:
@mattyr … sorry for the delay, will revert the changes I mentioned above

LeFnord:
@mattyr please can you check if it also happen with this PR https://github.com/ruby-grape/grape-swagger/pull/597

mattyr:
looks like #597 does fix it.  will close this pull request.

PR #593, opened 2017-03-07
clarify hidden option in readme

thogg4:


README.md
```diff
@@ -438,10 +438,10 @@ desc "Return super-secret information", {
 <a name="hiding" />
 #### Hiding an Endpoint
 
-You can hide an endpoint by adding ```hidden: true``` in the description of the endpoint:
+You can hide an endpoint by adding ```hidden: true``` on the verb method of the endpoint, such as `get`, `post` and `put`:
```
LeFnord: @thogg4 … thanks for contribution …
please don't replace existed documentation, it is still correct, only add as additional possibility to hide an endpoint

thogg4:
Hey @LeFnord I couldn't get the existing example to work, but as requested I added it back. The readme now has both.

LeFnord:
thanks @thogg4 … yeah the example … what I mean is this version to hide an endpoint:
```ruby
desc 'put specific foo', hidden: true
put ':id' do
  # your code goes here
end
```
and your version of it – this is new to me 😉 
```ruby
patch ':id', hidden:true do
  # your code goes here
end
```

Issue #590, opened 2017-03-01
Ruby 2.3 - frozen string literal compatibility issues


mattyr:
i originally mentioned in pr #589 , but there are rubocop failures if you run ```bundle exec rake``` under ruby 2.3.

the source of the issue is rubocop's [FrozenStringLiteralComment](http://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/FrozenStringLiteralComment) check, which fails if missing the comment in source code files (which is the case for a lot of the files in the project).

the issue is that globally adding those comments introduces bugs, ie:

```
1) GrapeSwagger::DocMethods::OperationId build :get 
     Failure/Error: let(:route) { Grape::Router::Route.new(method, '/path', requirements: {}) }
     
     RuntimeError:
       can't modify frozen String
     # /Users/matt/.rvm/gems/ruby-2.3.3@grape-swagger/gems/grape-0.19.1/lib/grape/router/pattern.rb:39:in `build_path'
     # /Users/matt/.rvm/gems/ruby-2.3.3@grape-swagger/gems/grape-0.19.1/lib/grape/router/pattern.rb:19:in `initialize'
     # /Users/matt/.rvm/gems/ruby-2.3.3@grape-swagger/gems/grape-0.19.1/lib/grape/router/route.rb:67:in `new'
     # /Users/matt/.rvm/gems/ruby-2.3.3@grape-swagger/gems/grape-0.19.1/lib/grape/router/route.rb:67:in `initialize'
     # ./spec/lib/operation_id_spec.rb:11:in `new'
     # ./spec/lib/operation_id_spec.rb:11:in `block (3 levels) in <top (required)>'
     # ./spec/lib/operation_id_spec.rb:23:in `block (4 levels) in <top (required)>'
```

the source of the issue isn't always grape-swagger, but as in the example above, grape itself

seems to me there are a few directions to go to "fix" this:

- make the referenced gems, and grape-swagger, compatible with frozen strings
- globally ignore the FrozenStringLiteralComment in rubocop.yml
- selectively include the frozen string literal comment in grape-swagger's files, and add specific ignores in rubocop.yml
- don't support ruby >= 2.3

not a fan of the last option, seems like the first option is best but might take some coordination.

(btw -- the only errors were generated from spec/lib/operation_id_spec.rb, but i'm not sure if that's the tip of an iceberg).

thoughts?

mattyr:
[...]

PR #595, opened 2017-03-08
issue #590: fixes rubocop offences

LeFnord:


spec/lib/operation_id_spec.rb
```diff
@@ -1,3 +1,4 @@
+# frozen_string_literal: false
```
mattyr: ah, this is how you got it to pass.  is there any desire to address the frozen string literal issue upstream in grape, so that this can be set to "true"?

LeFnord: not as I know, but feel always welcome to improve it

Issue #594, opened 2017-03-07
Grape deprecated passing options hash to desc 


thogg4:
It is now deprecated to pass an options hash to `desc` in grape like this:

```
desc 'get an object', hidden: true
```
As shown here:
https://github.com/ruby-grape/grape/blob/master/lib/grape/dsl/desc.rb#L50

And (as far as i can tell) passing options like that no longer has an effect when using the latest versions of grape and grape-swagger.

For example, that line above doesn't seem to hide the endpoint.

The grape project seems to prefer setting options like this now:

```
desc 'get an object'
route_setting :swagger, hidden: true
get '/object/:id' do

end
```

As described here:
https://github.com/ruby-grape/grape#describing-and-inspecting-an-api

Am I missing something?
Would you be interested in a pull request that takes this into account?

LeFnord:
heap , this warning exists since grape 0.15.0, but it doesn't mean one can not provide options, this can be done in a block …
```ruby
desc 'get an object' do
  hidden true
end
```

thogg4:
@LeFnord that example was the first thing I tried. It gives me this error:

`NoMethodError: undefined method `hidden' for #<#<Class:0x007fbb1464fad8>:0x007fbb1464f150>`

LeFnord:
it is a bit curious, cause there is a spec (https://github.com/ruby-grape/grape-swagger/blob/master/spec/swagger_v2/hide_api_spec.rb) which exactly uses your first version, without  any warnings → have a look on the travis builds

the the second version raises an error, sorry I forget, that all methods of the block must be defined for grape, that is why I personally prefer the hash style, it is much more flexible and more independent form grape

by the way, have a look on the specs, the hash style is very often used, without any problems

edgarjs:
Does this cover the `is_array: true` option for the endpoint? For what I can see it only fixes the `hidden` one.

thogg4:
@edgarjs this specific issue was addressed by #596 it doesn't cover `is_array`

PR #596, opened 2017-03-08
use route_settings for hidden and operations extensions

thogg4:
@LeFnord This pr may be a little heavy handed, but I wanted to show you what I meant by having grape-swagger use the preferred way to pass route settings.

Please let me know what you think and what I can improve and change.

Addresses https://github.com/ruby-grape/grape-swagger/issues/594


LeFnord:
will give it a try and a deeper look on the weekend, ok?

thogg4:
@LeFnord what do you think about this approach?

grape-swagger.gemspec
```diff
@@ -14,7 +14,7 @@ Gem::Specification.new do |s|
   s.license     = 'MIT'
 
   s.required_ruby_version = '>= 2.2.6'
-  s.add_runtime_dependency 'grape', '>= 0.16.2'
+  s.add_runtime_dependency 'grape', '>= 0.19.1'
```
LeFnord: please revert it back

lib/grape-swagger/endpoint.rb
```diff
@@ -337,7 +337,8 @@ def model_name(name)
     end
 
     def hidden?(route, options)
-      route_hidden = route.options[:hidden]
+      route_hidden = route.settings.try(:[], :swagger).try(:[], :hidden)
+      route_hidden = route.options[:hidden] if route.options.key?(:hidden)
```
LeFnord: am I right, the original behavior still available?

spec/swagger_v2/api_swagger_v2_extensions_spec.rb
```diff
@@ -19,10 +19,11 @@ class ExtensionsApi < Grape::API
           { 'declared_params' => declared(params) }
         end
 
+        route_setting :x_operation, some: 'stuff'
```
LeFnord: I like it, makes the extensions more consistent, thanks

lib/grape-swagger/endpoint.rb
thogg4: Yeah, the original behavior is default.

LeFnord:
👍 @thogg4 please add the changelog entry, then I'll mörge it

PR #471, opened 2016-07-08
Allow Security Definitions Objects to be defined

bendodd:
Add securityDefinitions to allow Swagger UI to authenticate

A basic solution for #432

PR #602, opened 2017-04-13
Allow Security Object to be defined

markevich:
Hi!
It's relates to #471 
We can describe ```securityDefinition```, but it is only the definition :) We need to define second object ```security``` to make swagger add authorization key to every request.


cberkom:
Any idea when this will be rolled into a new gem release?

PR #607, opened 2017-05-04
Allow body parameter name to be specified

tjwp:
For deeply nested resources, the generated name for a body parameter, based on the operation, can get very long. This change allows the generated name to be overridden using a `:body_name` option on the desc.

While adding tests for this, I found that the relevant specs used a `Grape::Request` while the actual object that is passed is a `Grape::Router::Route`, so they are updated accordingly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 #### Features
 
 * Your contribution here.
+* [#607](https://github.com/ruby-grape/grape-swagger/pull/607): Allow body parameter name to be specified - [@tjwp](https://github.com/tjwp).
 
 #### Fixes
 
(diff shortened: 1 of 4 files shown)


lib/grape-swagger/doc_methods/move_params.rb
```diff
@@ -42,9 +42,9 @@ def parent_definition_of_params(params, path, route)
 
           move_params_to_new(definition, params)
 
-          definition[:description] = route.description if route.respond_to?(:description)
+          definition[:description] = route.description if route.try(:description).present?
```
LeFnord: think the `present?` isn't needed here

CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 #### Features
 
-* Your contribution here.
```
LeFnord: please re-add it

tjwp:
@LeFnord feedback addressed. Thanks for the quick review!

LeFnord:
thanks for the quick response 😄

PR #608, opened 2017-05-08
support extensions on the root object

thogg4:


thogg4:
@LeFnord great point. how is that?

LeFnord:
@thogg4 please fix the rubocop stuff and update the PR, so that it can be mörged 😄

PR #613, opened 2017-05-30
Fix Proc with arity one in param values

timothysu:
Recently ran into an issue where having custom validations in `values` would cause documentation generation to fail.  This simply mitigates an error from occurring.

---

From https://github.com/ruby-grape/grape/blob/master/README.md :

Alternatively, a Proc with arity one (i.e. taking one argument) can be used to explicitly validate
each parameter value.  In that case, the Proc is expected to return a truthy value if the parameter
value is valid.

```ruby
params do
  requires :number, type: Integer, values: ->(v) { v.even? && v < 25 }
end
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 #### Features
 
 * Your contribution here.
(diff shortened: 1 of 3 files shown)


LeFnord:
thanks @timothysu and sorry for the late mörge

PR #618, opened 2017-07-10
Relax the minor version requirement

joneslee85:
Ruby 2.2 series is pretty much alive and many production apps are still not upgraded to last minor version.

Locking to 2.2.6 would make things much more difficult for users to make transition.


LeFnord:
they apps should be upgraded to latest patch version, not minor, especially if they run in production

… on how many gems you did? think on, in this time you upgraded your ruby 😉

PR #622, opened 2017-07-19
Add support for 'brackets' collection format

korstiaan:
Although not in the official spec, it is support by `swagger-ui`. The `multi` format (`foo=a&foo=b`) doesn't work with `rack` base apps (like `rails`). Rack wants it to look like `foo[]a=&foo[]=b`, which is exactly what the `brackets` format specifies.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 #### Features
 
-* Your contribution here.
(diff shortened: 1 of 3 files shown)


LeFnord:
thanks @korstiaan

viral810:
@LeFnord Am I missing anything here? seems like this doesn't do anything except the fact that you can use `collectionFormat: 'brackets'` in documentation hash. 

For example, 
```
if param_type_is_array?(param_type)
            options[:is_array] = true
            name += '[]' if array_use_braces? # Shouldn't this also check the collectionFormat from the documentation?
          end
```

PR #631, opened 2017-09-11
Fix order of mounts with overrides

adie:
The problem is that `grape` executes endpoints in the order of definition (like defined first - matched first), but documentation was showing information from the last defined endpoints.
Not sure if it's issue with the `grape` itself, but I think `grape-swagger` should be consistent with current state of `grape` gem.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 #### Fixes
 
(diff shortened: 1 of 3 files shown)


LeFnord:
thanks @adie … think it isn't really a 'problem' cause the documentation itself is a JSON and the the path section is a kind of a hash, so it has no order 😉 … but yeah, why not

Issue #626, opened 2017-07-25
Cannot post Array params


michaelize:
```
params do
  ...
  optional :user_profiles, type: Array, default: [] do
    requires :value, type: String, desc: "Value of the user profile"
  end
end```

Using the above gives me curl -X POST --header 'Content-Type: application/x-www-form-urlencoded' --header 'Accept: application/json' -d 'user_profiles[value]=1' 'http://localhost:3000/api/v1/accounts/3/users', and it rejects with user_profiles invalid.

Can someone tell me what am I doing wrong here? Thx!

michaelize:
If I run this curl command it works: `curl -X POST --header 'Content-Type: application/x-www-form-urlencoded' --header 'Accept: application/json' -d 'user_profiles[][value]=1' 'http://localhost:3000/api/v1/accounts/3/users'`

This seems to be a bug. Should be the same in [#617](https://github.com/ruby-grape/grape-swagger/issues/617)

Issue #617, opened 2017-07-04
Cannot get Array of Hash working


ahmeij:
Thanks for building this great bridge between swagger and grape.

I have spent a good bit of time trying to get the following to work, I have even created a minor rails test environment to see if the complicated setup is the cause but I can reproduce it with only a few lines of code.

I'd like to create an order with multiple order lines. 

```ruby
require 'grape-swagger'

module Twitter
  class API < Grape::API
    format :json
    prefix :api

    resource :orders do
      desc 'Create an order'
      params do
        optional :order_type
        optional :order_lines, type: Array do
          optional :product
          optional :quantity, type: Integer
        end
      end
      post do
        params
      end
    end

    add_swagger_documentation
  end
end
```

This gives me the following swagger-ui result

<img width="961" alt="screen shot 2017-07-04 at 15 28 49" src="https://user-images.githubusercontent.com/527143/27832061-883f8d66-60cd-11e7-8214-b30bf3bd07b6.png">

I was really hoping it would allow for these nested order-lines to work. However checking the curl line the posted output is not as expected:

I'd expect it to post order_lines[0][product]='cheese'&order_lines[0][quantity]=1 

Also I think the UI looks funky, but that might be just swagger?

I'm really at a loss here if this is a grape-swagger, grape or swagger-ui issue, I am hoping you guys know what is wrong with my example.

michaelize:
Did you ever get it solved?

ahmeij:
No unfortunately not :(

michaelize:
Sounds like a bug, cause if I run a curl command, it works without any issues.

seguidor777:
I have the same issue, array of JSON is not properly shown in the UI

TheAbyss:
Please address +1, Have a similar issue after my frontend developer pointed out he could now see the array params.

dblock:
Please contribute @TheAbyss.

QETHAN:
+1

stephancom:
+1

MiralDesai:
+1

VoTaiTri:
+1

EmCousin:
+1

mfbmina:
+1

FUT:
+1

rabendrasharma:
Please add this feature. Any workaround on this.

QxOxOxQ:
+1

jacek213:
could the maintainer of this package speak his mind about this?

[...]

PR #637, opened 2017-09-27
Add an option to add braces to array params

### LINKED ISSUE #636: Array group parameters
Hi, I have an issue that I believe was fixed by #221 but then got overridden by #455.
I think it was needed to support both cases, not just "fix" those specs: https://github.com/ruby-grape/grape-swagger/pull/455/files#diff-2b857d0b8829328970ec856ccd7d8ac3L43

Right now things like
```ruby
requires :group, type: Array do
  requires :param_1, type: Integer
  requires :param_2, type: String
end
```
are being displayed as a simple `Hash`, not as an `Array`.

Even more, for some reason it applies `Array` type to the nested parameters:
swagger-group-issue
I would expect to have `group[][param_1]` with type `integer` instead

adie:
Right now it's more like a proof of concept in attempt to fix #636 
I've added `array_use_braces` option for the `add_swagger_documentation` so that it would be a feature toggle. Not sure if it's ok to do this globally.
It's possible that it fixes other issues with array params like #617 and #626 
Please let me know what do you think of such approach

TODO:
* [x] Distinguish between `water` and `watermelon`, e.g. check that it does not simply start from the string, but matches whole name.
* [ ] Document new option in README

### DIFF
--- a/lib/grape-swagger/endpoint.rb
+++ b/lib/grape-swagger/endpoint.rb
@@ -1,7 +1,8 @@
 # frozen_string_literal: true
 
 require 'active_support'
-require 'active_support/core_ext/string/inflections.rb'
+require 'active_support/core_ext/string/inflections'
+require 'grape-swagger/endpoint/params_parser'
 
 module Grape
   class Endpoint
@@ -265,7 +266,7 @@ module Grape
       default_type(required)
 
       request_params = unless declared_params.nil? && route.headers.nil?
-                         parse_request_params(required, settings)
+                         GrapeSwagger::Endpoint::ParamsParser.parse_request_params(required, settings)
                        end || {}
 
       request_params.empty? ? required : request_params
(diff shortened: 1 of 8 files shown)


lib/grape-swagger/endpoint.rb
```diff
@@ -282,14 +282,25 @@ def default_type(params)
...
+    def parse_request_params(params, settings)
+      array_keys = []
       params.select { |param| public_parameter?(param) }.each_with_object({}) do |param, memo|
         name, options = *param
         param_type = options[:type]
         param_type = param_type.to_s unless param_type.nil?
-        array_key = name.to_s if param_type_is_array?(param_type)
-        options[:is_array] = true if array_key && name.start_with?(array_key)
+
+        array_keys << name.to_s if param_type_is_array?(param_type)
```
adie: Previously it was overriting `array_key` which could lead to the problem with such nesting:
```ruby
optional :a do
  optional :b do
    optional :c
  end
  optional :d
end
```
Previously here `:d` would not be detected as an array type

lib/grape-swagger/endpoint.rb
```diff
@@ -282,14 +282,25 @@ def default_type(params)
...
+
+        keys = array_keys.find_all { |key| name.start_with? key }
+        if keys.any?
+          options[:is_array] = true
+          if settings[:array_use_braces] && !(options[:documentation] && options[:documentation][:param_type] == 'body')
+            keys.sort.reverse_each do |key|
```
adie: I'm doing sort and reverse each here so that we will start replacing with longer strings, like with `keys == ['a[b]', 'a']`: `a[b][c]` => `a[b][][c]` and only then => `a[][b][][c]`. Otherwise it will replace it to `a[][b][c]` and then wouldn't find `a[b]` to replace

LeFnord:
@adie any progress on it?

LeFnord:
@adie … any work on it, I realy like to have it, cause it solves some bugs …   
or can somebody it take over?

brianweiner:
@LeFnord I've been using this locally for an API we're developing for our integrating partners -- what steps do you need me to take to get this PR merge-worthy?

LeFnord:
Hi @brianweiner … please fulfil your task list → document the new option, and of course, resolve the conflicts, then I'm realy happy to mörge it.

thanks 😄

brianweiner:
https://github.com/ruby-grape/grape-swagger/pull/695

LeFnord:
closed in favour og #695

PR #641, opened 2017-10-17
Default success codes to be excluded when http_codes define one already

anakinj:
It was not possible to override a POSTs default 201 status code. This change will prevent the default success code to be added if the http_codes array already include a success code.

### DIFF
--- a/spec/swagger_v2/api_swagger_v2_status_codes_spec.rb
+++ b/spec/swagger_v2/api_swagger_v2_status_codes_spec.rb
@@ -1,0 +1,52 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe 'http status code behaivours' do
+  include_context "#{MODEL_PARSER} swagger example"
+
+  subject do
+    get '/swagger_doc'
+    JSON.parse(last_response.body)
+  end
+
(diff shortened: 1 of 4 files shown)


spec/swagger_v2/api_swagger_v2_status_codes_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+  subject do
+    get '/swagger_doc'
+    JSON.parse(last_response.body)
+  end
+
+  context 'when non-default success codes are deifined' do
+    let(:app) do
+      Class.new(Grape::API) do
+        desc 'Has explicit http_codes defined' do
+          http_codes [{ code: 202, message: 'We got it!' },
```
LeFnord: what's happened if we have mixed codes, something like this:
```ruby
http_codes [
  { code: 200, message: 'all fine' },
  { code: 404, message: 'Not found' }
]
```
maybe adding a spec to document the behaviour

anakinj: I've added some more status codes combinations to the specs

LeFnord:
thanks @anakinj

PR #654, opened 2017-12-05
consumes support for PATCH

anakinj:
Trying to get our Swagger docs to pass some validations and noticed that for some reason the consumes setting was not supported for PATCH methods. This should fix the issue.


LeFnord:
thanks @anakinj

PR #667, opened 2018-02-24
Make route summary optional

### LINKED ISSUE #661: Removing duplicate fields
Currently if an API endpoint only has a desc (which is fairly common), grape-swagger uses that description for the summary, description, and response description. To me, that seems like duplication, and better would be to not set the summary unless it is specifically defined

Do people agree / disagree and why?

If it was agreed, I would be happy to do the PR, but wanted to check before putting in the effort.

obduk:
Resolves #661

Currently if no summary is provided, grape-swagger uses the description.
This creates needless duplicate fields. As summary is an optional field
this changes functionality so summary is only populated if specifically
needed.

Reference https://swagger.io/specification/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 #### Features
 
-* Your contribution here.
+* [#667](https://github.com/ruby-grape/grape-swagger/pull/667):  Make route summary optional - [@obduk](https://github.com/obduk).
 
 #### Fixes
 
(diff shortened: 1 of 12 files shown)


LeFnord:
thanks @obduk … it would be nice, to have some specs with summary, to document the behaviour,  
that it can be explitely set, such as the PR name suggest

obduk:
@LeFnord thanks for looking at this. There are already tests for this in spec/swagger_v2/namespaced_api_spec.rb as it is not new functionality.

LeFnord:
ok, then I'm fine with it and mörge it, thanks again

PR #669, opened 2018-03-06
Fix handling of HTTP status codes from routes

milgner:
After updating Grape, the documentation could not be generated anymore.

Upon investigation it turned out that the HTTP codes were not stored in a hash (anymore) but as a two-element array with the status code first and a message second. After applying this patch, documentation was generated successfully again.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -21,7 +21,7 @@ Metrics/AbcSize:
 # Offense count: 3
 # Configuration parameters: CountComments.
 Metrics/ClassLength:
-  Max: 283
+  Max: 287
 
 # Offense count: 10
 Metrics/CyclomaticComplexity:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 #### Fixes
 
 * [#664](https://github.com/ruby-grape/grape-swagger/pull/662): Removed all references to obsolete `hide_format` parameter - [@jonmchan](https://github.com/jonmchan).
+* [#669](https://github.com/ruby-grape/grape-swagger/pull/669): Fix handling of http status codes from routes - [@milgner](https://github.com/milgner).
 
 ### 0.28.0 (February 3, 2018)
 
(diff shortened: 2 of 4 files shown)


milgner:
Damn, did this drive-by-change and forgot to adapt the specs... Will do so...

milgner:
After additional investigation it looks like there are multiple ways to define status codes and they result in different representations in the internal structures.

The routes used for the specs in this gem use

```
          http_codes: [
            { code: 200, message: 'get Splines', model: Api::Entities::Splines },
            { code: 422, message: 'SplinesOutError' }
          ]
```

whereas the Grape documentation (and our application) uses

```
  failure [[401, 'Unauthorized', 'Entities::Error']]
```

which also ends up in the same structure.

After checking the Grape README, I did not explicitly see `http_codes` used. But in the end it seems like `grape-swagger` will need to support both versions. So I'll have to adapt the code to deal with both constellations.

LeFnord:
the keys has changed, see: [grape#describing-methods](https://github.com/ruby-grape/grape#describing-methods), so former `entity` is `success` and `http_codes`are4 `failure` (should also be plural imho), bugt internaly is no difference

the difference is, that `failure` can accept an Array of Arrays in grape and in grape-swagger also an Array of Hashes

please have in mind these keys would only be used by grape-swagger

milgner:
Just a quick update: I haven't forgotten about this PR, just been a bit low on time for OSS work - I have a couple of free days coming up this week and will take care of getting all the different variations of this covered then.

LeFnord:
@milgner thanks for the update 😄

milgner:
Oh no, looks like my changes tipped the metrics for `Metrics/ClassLength` of `Endpoint` over the edge... :see_no_evil:

LeFnord:
thanks @milgner … please can you rebase and update the CHANGELOG with your entry

milgner:
:+1: Done!

LeFnord:
thanks again …

PR #670, opened 2018-03-16
Add support for deprecated field

ioanatia:
`deprecated` is one of the fixed fields in an `Operation` object.

It's included in all versions:
- https://github.com/OAI/OpenAPI-Specification/blob/master/versions/1.2.md#523-operation-object
- https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#operation-object
- https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#operation-object

### DIFF
--- a/lib/grape-swagger/endpoint.rb
+++ b/lib/grape-swagger/endpoint.rb
@@ -123,9 +123,14 @@ module Grape
       method[:responses]   = response_object(route)
       method[:tags]        = route.options.fetch(:tags, tag_object(route, path))
       method[:operationId] = GrapeSwagger::DocMethods::OperationId.build(route, path)
+      method[:deprecated] = deprecated_object(route)
       method.delete_if { |_, value| value.blank? }
 
       [route.request_method.downcase.to_sym, method]
+    end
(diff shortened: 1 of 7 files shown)


lib/grape-swagger/endpoint.rb
```diff
@@ -4,6 +4,7 @@
 require 'active_support/core_ext/string/inflections.rb'
 
 module Grape
+  # rubocop:disable Metrics/ClassLength
```
LeFnord: please don't add it here, specify it in the .rubocop-todo.yml

LeFnord:
thanks @ioanatia …

lib/grape-swagger/endpoint.rb
ioanatia: okay, and I will fix the other two rubocop issues that causes the CI to fail.

LeFnord:
thanks again @ioanatia

ioanatia:
thanks for reviewing this so quickly :smile_cat:

PR #673, opened 2018-04-01
Update README.md - usage of `deprecated` fix field

ioanatia:
This is a follow up on https://github.com/ruby-grape/grape-swagger/pull/670 where support for the `deprecated` fix field was added, but the docs were not updated :sweat_smile: .


LeFnord:
thanks @ioanatia …

Issue #680, opened 2018-07-12
Error schemas are lost for all responses when an endpoint can return a 204


adstratm:
If a Grape Endpoint returns a 204 response when successful, the schemas are lost for any of the error responses as well.  For the following mock endpoint, 

```
desc 'delete something',
     is_array: true,
     success: { status: 204 },
     failure: [
       [401, 'Unauthorized', Entities::ApiError],
       [403, 'Forbidden', Entities::ApiError],
       [404, 'Not Found', Entities::ApiError]
     ]

delete do
  status 204
end
```

the generated docs from `0.29` are as follows: 

```
{"delete"=>
  {"description"=>"delete something",
   "produces"=>["application/json"],
   "responses"=>
    {"204"=>{"description"=>"delete something"},
     "401"=>{"description"=>"Unauthorized"},
     "403"=>{"description"=>"Forbidden"},
     "404"=>{"description"=>"Not Found"}},
   "tags"=>["schema_lost"],
   "operationId"=>"deleteSchemaLost"}}
```

But because a schema was provided for the 400-level errors, this information should be reflected in the generated doc:

```
{"delete"=>
  {"description"=>"delete something",
   "produces"=>["application/json"],
   "responses"=>
    {"204"=>{"description"=>"delete something"},
     "401"=>
      {"description"=>"Unauthorized",
       "schema"=>{"$ref"=>"#/definitions/ApiError"}},
     "403"=>
      {"description"=>"Forbidden",
       "schema"=>{"$ref"=>"#/definitions/ApiError"}},
     "404"=>
      {"description"=>"Not Found",
       "schema"=>{"$ref"=>"#/definitions/ApiError"}}},
   "tags"=>["schema_lost"],
   "operationId"=>"deleteSchemaLost"}}
```

The issue is that any responses belonging to an endpoint that *could* respond with 204 are discarded.  I'll open a pull request momentarily with the fix.

adstratm:
[Pull request](https://github.com/ruby-grape/grape-swagger/pull/681)

adstratm:
Merged with a73657b24d84231c9e418f54f575744c4460e3a5.  Thanks!

PR #681, opened 2018-07-12
Provide error schemas when an endpoint can return a 204

adstratm:
See [issue](https://github.com/ruby-grape/grape-swagger/issues/680)


LeFnord:
thanks for fixing @adstratm

PR #684, opened 2018-07-19
Add response header OpenAPI spec generation

jdmurphy:
Add support to expose [response headers object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#headers-object)

* Available only with desc DSL block syntax under success and failure keys.
* Consumer is responsible for documenting headers

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -16,6 +16,9 @@ Metrics/BlockLength:
   Exclude:
     - spec/**/*
 
+Metrics/ClassLength:
+  Max: 300
+
 Metrics/LineLength:
   Max: 120
   Exclude:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 #### Features
 
-* Your contribution here.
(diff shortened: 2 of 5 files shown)


LeFnord:
@jdmurphy thanks for it, but I'm a bit confused, your title says something abvout response example, but your code handles headers …

background: there exists a key `headers` for the describe method, for documenting it  
did your change take the same line?

… and sorry, mörged your second PR rivght now

jdmurphy:
@LeFnord, sorry that was some bad copy/pasting because the feature is similar in scope to @gamartin's addition of response examples.

When I used the `headers` key in the describe block, it showed up as a parameter and not in the response area. It is possible I was not using it correctly/there was something I missed when I added it; if so please let me know.

LeFnord:
@jdmurphy … yeap, you are right, `success` and `failures` will be used for documenting the response and the `headers` key as kind as parameter for the request

LeFnord:
thanks for your contributions @jdmurphy

PR #688, opened 2018-07-23
Deep merge params for definition

jdmurphy:
I found that when using body params in PUT and POST, a nested entity in the params that has a description set at the entity level will not have that description in the OpenAPI spec output. For example given:
```
params do
  requires :spline, description: 'It is a spline', documentation: {param_type: 'body'}, type: Hash do
    requires :x, type: Numeric
    requires :y, type: Numeric
  end
  optional :reticulated, type: Boolean, default: true, desc: 'True if the spline is reticulated.'
end
```
a definition will be created (notice that the `spline` hash under properties has no description): 
```
"postSplines":{  
   "type":"object",
   "properties":{  
      "spline":{  
         "type":"object",
         "properties":{  
            "x":{  
               "type":"integer",
               "format":"int64"
            },
            "y":{  
               "type":"integer",
               "format":"int64"
            }
         },
         "required":[  
            "x",
            "y"
         ]
      },
      "reticulated":{  
         "type":"boolean",
         "description":"True if the spline is reticulated."
      }
   },
   "required":[  
      "spline"
   ],
   "description":"Create a spline."
 }
```

With my fix a deep merge occurs which maintains elements on the parent object when merging a nested element: (and the `description` of "It is a spline" is kept)
```
"postSplines":{  
   "type":"object",
   "properties":{  
      "spline":{  
         "type":"object",
         "description":"It is a spline",
         "properties":{  
            "x":{  
               "type":"integer",
               "format":"int64"
            },
            "y":{  
               "type":"integer",
               "format":"int64"
            }
         },
         "required":[  
            "x",
            "y"
         ]
      },
      "reticulated":{  
         "type":"boolean",
         "description":"True if the spline is reticulated."
      }
   },
   "required":[  
      "spline"
   ],
   "description":"Create a spline."
}
```

while looking into this I noticed that the `recursive_call` function was using the `property` parameter as a key without regard to its format which could lead to having string and symbols mixed together specifically when nested params are being evaluated (the nested param would be the string whereas every other element would be a symbol), so I added a call to `.to_sym` to ensure use of symbols throughout since we are using Hash and not HashWithIndifferentAccess.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 #### Fixes
 
+* [#688](https://github.com/ruby-grape/grape-swagger/pull/688): Use deep merge for nested parameter definitions - [@jdmurphy](https://github.com/jdmurphy).
 * Your contribution here.
 
 ###  0.30.1 (July 19, 2018)
--- a/lib/grape-swagger/doc_methods/move_params.rb
+++ b/lib/grape-swagger/doc_methods/move_params.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'active_support/core_ext/hash/deep_merge'
+
 module GrapeSwagger
   module DocMethods
     class MoveParams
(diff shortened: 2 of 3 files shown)


LeFnord:
thanks @jdmurphy

PR #691, opened 2018-07-30
Fix is_array check for non-contiguous sub and parent entities

jdmurphy:
While reviewing an API made with Grape, I noticed that an entity created for a POST request param body did not have the correct markings for array and object. In digging into the code, I found that the `parse_request_params` was dependent on the ordering of the params coming in. Specifically, since we initialize `array_key` [outside the loop](https://github.com/ruby-grape/grape-swagger/blob/v0.30.1/lib/grape-swagger/endpoint.rb#L309), and then set it [below](https://github.com/ruby-grape/grape-swagger/blob/v0.30.1/lib/grape-swagger/endpoint.rb#L314), it _should_ exist for elements of an array, but I found that my params were in an order that caused sub elements to not be by their parents. For example (from tests so that I don't have to leak my API internals):
```
[
  ['id', { required: true, type: 'String' }],
  ['description', { required: false, type: 'String' }],
  ['stuffs', { required: true, type: 'Array' }],
  ['stuffs[owners]', { required: true, type: 'Array' }],
  ['stuffs[creators]', { required: true, type: 'Array' }],
  ['stuffs[owners][id]', { required: true, type: 'String' }],
  ['stuffs[creators][id]', { required: true, type: 'String' }]
]
```
would be updated to (notice `'stuffs[owners][id]'`).
```
[
  ['id', { required: true, type: 'String' }],
  ['description', { required: false, type: 'String' }],
  ['stuffs', { required: true, type: 'Array', is_array: true }],
  ['stuffs[owners]', { required: true, type: 'Array', is_array: true }],
  ['stuffs[creators]', { required: true, type: 'Array', is_array: true }],
  ['stuffs[owners][id]', { required: true, type: 'String' }],
  ['stuffs[creators][id]', { required: true, type: 'String', is_array: true }]
]
```

I updated parse_request_params to keep all array keys and then see if the name of current param matches any of them exactly or matches any of the array keys with a `[` appended which would indicate it is a sub element.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 #### Fixes
 
 * [#688](https://github.com/ruby-grape/grape-swagger/pull/688): Use deep merge for nested parameter definitions - [@jdmurphy](https://github.com/jdmurphy).
+* [#691](https://github.com/ruby-grape/grape-swagger/pull/691): Disregard order when parsing request params for arrays - [@jdmurphy](https://github.com/jdmurphy).
 * Your contribution here.
 
 ###  0.30.1 (July 19, 2018)
--- a/lib/grape-swagger/endpoint.rb
+++ b/lib/grape-swagger/endpoint.rb
@@ -306,13 +306,13 @@ module Grape
     end
 
     def parse_request_params(params)
-      array_key = nil
+      array_keys = []
       params.select { |param| public_parameter?(param) }.each_with_object({}) do |param, memo|
         name, options = *param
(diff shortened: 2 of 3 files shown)


LeFnord:
thanks again @jdmurphy 😄 …

jdmurphy:
@LeFnord, any chance we could get a release since there are a few fixes lined up since the last one?

LeFnord:
yeap sure, will cut a new release the next days … 😄

PR #692, opened 2018-08-01
Fix compatibility description in readme

darren987469:
Change grape-entity compatibility description from `=> 0.5.0` to `>= 0.5.0`.


LeFnord:
a CHANGELOG entry isn't needed here, so I removed it by resolving the conflicts, will mörge it after tests passes

LeFnord:
thanks @darren987469

PR #701, opened 2018-09-01
Default in MoveParams#property_keys

onemanstartup:
Basically the same PR as https://github.com/ruby-grape/grape-swagger/pull/630 only different key.

default property is already supported, but not when moved from params to definition

### DIFF
--- a/lib/grape-swagger/doc_methods/move_params.rb
+++ b/lib/grape-swagger/doc_methods/move_params.rb
@@ -207,7 +207,7 @@ module GrapeSwagger
         end
 
         def property_keys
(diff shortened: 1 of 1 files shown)


LeFnord:
Hi @onemanstartup … a spec as kind of documentation would be nice 😉 … thanks

onemanstartup:
@LeFnord There is no specs for previous PR... 😄 
Do you want me to add those specs here https://github.com/ruby-grape/grape-swagger/blob/master/spec/lib/move_params_spec.rb ?

onemanstartup:
@LeFnord thanks

PR #702, opened 2018-09-10
Use correct parameter name in docs (array_use_braces).

dslh:
This is just a quick PR to ensure the new `array_use_braces` parameter is named correctly in the README.

Let me know if you'd like a changelog entry added, I didn't think it was merited in this case.

Thanks to @adie and @brianweiner for the `array_use_braces` feature, and thanks to all contributors and maintainers 🙂


LeFnord:
thanks @dslh

PR #706, opened 2018-09-12
DSL::Desc hash options are overriden by initializing them in block

ysv:


spec/swagger_v2/nicknamed_api_spec.rb
```diff
@@ -3,9 +3,17 @@
...
+    expose :at do
+      Time.now
+    end
+  end
+
   def app
     Class.new(Grape::API) do
-      desc 'Show this endpoint', nickname: 'simple'
+      desc 'Show this endpoint', nickname: 'simple' do
+        success Test
```
ysv: success overrides nickname so nickname is set to default value now

dblock:
We should get this test merged after a passing build. Want to clean it up @ysv?

PR #713, opened 2018-09-29
Reimplemented api_documentation and specific_api_documentation options.

### LINKED ISSUE #710: What did we do to api_documentation and specific_api_documentation
We used to have `specific_api_documentation` and `api_documentation` that allowed merging anything onto a Grape endpoint. That seems to be gone, although https://github.com/ruby-grape/grape-swagger/blob/c00c5b40c5ab809047aa40bc36dcd9b871ad24ef/lib/grape-swagger/doc_methods.rb#L100 still remains. 

We used it to add custom fields like our own `role_open` in this example:

```ruby
      add_swagger_documentation \
        doc_version: 'v1',
        api_documentation: { desc: 'Artsy API v1 documentation.', role_open: true },
        specific_api_documentation: { desc: 'Artsy API v1 documentation for a specific API.', role_open: true },
        hide_format: true
```

It would be accessible in `env['api.endpoint'].route_setting(:description)[:role_open]`.

1) Is there a way to do this today?
2) Should the default value in the link above be deleted?

dblock:
Closes #710. 

We use this to allow open access to the swagger documentation by adding a block like this with a `role_open` option.

```ruby
add_swagger_documentation \
  api_documentation: { desc: 'Artsy API v1 documentation.', role_open: true },
  specific_api_documentation: { desc: 'Artsy API v1 documentation for a specific API.', role_open: true }
```

This is consumed by a middleware that looks roughly like this:

```ruby
class RoleMiddleware < Grape::Middleware::Base
  def before
    return if route_description[:role_open]
    return if authenticated?
    error!('Forbidden', 403, 'Not authorized')
  end
  
  private

  def route_description
    env['api.endpoint'].route_setting(:description) || {}
  end
        
  def authenticated?
     ... warden....
   end
end
```

LMK if you think this can/should be achieved differently.


dblock:
@LeFnord @Bugagazavr Anyone around?

dblock:
Maybe even @tim-vandecasteele? I don't feel like merging this myself without at least one of the existing maintainers pitching in a word ... but I guess I'll go ahead soon and try to push some PRs through.

tim-vandecasteele:
It's been a while since I showed my face here 😂 but I don't see any issue with this @dblock

dblock:
🙇

PR #717, opened 2018-10-23
Adds support for grape >= 1.2

myxoh:
This PR is required after: https://github.com/ruby-grape/grape/pull/1803
to make sure that `grape-swagger` is compatible with `grape` after the breaking changes introduced in `1.2.0`

Because of the nature of the changes in the grape gem, this version won't be compatible with prior versions of grape


myxoh:
Not sure how to handle the fact that this is dependant on the other code being merged. I've tested this locally manually using the branch and it was working

Gemfile
```diff
@@ -6,7 +6,7 @@ ruby RUBY_VERSION
 
 gemspec
 
-gem 'grape', case version = ENV['GRAPE_VERSION'] || '~> 1.0'
+gem 'grape', case version = ENV['GRAPE_VERSION'] || '~> 1.2'
```
dblock: Leave this at 1.0 for now, but add another entry to .travis.yml with the branch, so for example

```ruby
when 'INSTANCE'
 { git: '...', branch: '...' } # your own PR branch here
when 'HEAD'
  ...
```

UPGRADING.md
```diff
@@ -1,5 +1,9 @@
 ## Upgrading Grape-swagger
 
+### Upgrading to >= 0.32.0
+
+This version is *only* compatible with grape >= 1.2.0
```
dblock: Add some info on why, link to the other PR, add a link.

myxoh:
I've addressed the issue of making swagger compatible with previous versions of Grape.
I've also added a test that covers re-mounting if the version of grape is higher than 1.2

dblock:
This looks great. We can keep this moving once the Grape PR is merged.

dblock:
Looks like a rubocop failure. Lock it at 0.59.2 or upgrade to 0.60.0 and `rubocop -a ; rubocop --auto-gen-config`?

myxoh:
Thanks @dblock . Addressed.
By the way, I've sent you an e-mail regarding the request on the other thread. Let me know if you require anything else.

Gemfile
```diff
@@ -26,7 +26,7 @@ group :development, :test do
   gem 'rake'
   gem 'rdoc'
   gem 'rspec', '~> 3.0'
-  gem 'rubocop', '~> 0.59', require: false
+  gem 'rubocop', '~> 0.60', require: false
```
dblock: This is OK for now, but really should just be a specific version, because each new release (eg. 0.61) will produce new errors/warnings.

myxoh: Not sure whether it is desirable to lock the gem, rather than tackle the new errors / warnings as they come.
If you still feel I should lock this to a version, I'll do that

dblock: I got this in https://github.com/ruby-grape/grape-swagger/pull/719.

We lock rubocop because it's otherwise a moving target. Yesterday `~> 0.59` resolved to 0.59, but today it resolved to 0.60, producing new errors. So the same code built green yesterday but fails to build today, which is not desirable. 

New errors and warnings should totally be tackled, but not in the way of feature/fix work :)

CHANGELOG.md
```diff
@@ -1,12 +1,8 @@
-### 0.31.2 (Next)
+### 0.32.0 (Next)
 
 #### Features
 
-* Your contribution here.
```
LeFnord: please re-add it again, same for line 9

Gemfile
```diff
@@ -26,7 +26,7 @@ group :development, :test do
   gem 'rake'
   gem 'rdoc'
   gem 'rspec', '~> 3.0'
-  gem 'rubocop', '~> 0.59', require: false
+  gem 'rubocop', '0.60', require: false
```
LeFnord: please don't lock it, minor rubocop updates are much easier to handle then big ones

lib/grape-swagger/version.rb
```diff
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 module GrapeSwagger
-  VERSION = '0.31.2'
+  VERSION = '0.32.0'
```
LeFnord: revert, will be done on next release

LeFnord:
thanks @myxoh … please also add a travis entry

Gemfile
myxoh: @dblock Could you guys agree on a decision? I'm happy either way as I see both views as valid, but I don't want to step over either of you

dblock: Lets take it to #719

myxoh:
@LeFnord regarding adding a travis entry. I'm pressuming you mean against 1.12, however, that was not released yet, it's still on HEAD. Would you reckon we should wait until that's released and  then add it to the matrix?

LeFnord:
ok @myxoh …

LeFnord:
again … great work @myxoh … will mörge it, after specs are green 😄

PR #719, opened 2018-10-29
Upgraded and locked Rubocop at 0.60.0.

dblock:


Gemfile
```diff
@@ -26,7 +26,7 @@ group :development, :test do
   gem 'rake'
   gem 'rdoc'
   gem 'rspec', '~> 3.0'
-  gem 'rubocop', '~> 0.59', require: false
+  gem 'rubocop', '0.60.0', require: false
```
LeFnord: please, don't lock it

LeFnord:
@dblock … please revert the locking, then this branch will be mörged first, then the two others, thanks

dblock:
@LeFnord Thanks for a quick review. 

I think we should lock it. Yesterday's build with 0.59 worked, but today the same code fails to build because it's bringing a newer version. 

Does this ^ change anything? Let me know what you'd like me to do.

LeFnord:
@dblock … yeap agree with your arguments, but as I said, minor rubocop updates are easier to handle then big ones, and IMHO it is a maintainance task 😉 and should be done from time to time

dblock:
Rubocop doesn't quite follow semver, I changed to `~> 0.60.0`, LMK if that's acceptable?

dblock:
Marked HEAD to allow failures. Right now fails without #717

LeFnord:
mmh, please set it to `~> 0.60` … Batzov has its own idea of SemVer 😉 

yeap, that is fine, I'll mörge, so that #717 is a clean feature request, thanks again

PR #722, opened 2018-10-31
Refactor data type specs

lidimayra:
- Avoid explicit use of subject smell
[Reference](http://blog.davidchelimsky.net/blog/2012/05/13/spec-smell-explicit-use-of-subject/)

- Move stubbing behavior to inside the only block that really needs it.

- Use RSpec stub methods in order to specify return value instead of defining a whole new method with `.class_eval`
[Reference](https://relishapp.com/rspec/rspec-mocks/v/2-99/docs/method-stubs)

### DIFF
--- a/spec/lib/data_type_spec.rb
+++ b/spec/lib/data_type_spec.rb
@@ -3,15 +3,6 @@
 require 'spec_helper'
 
 describe GrapeSwagger::DocMethods::DataType do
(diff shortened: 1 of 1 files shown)


dblock:
Thanks for this.

lidimayra:
Sure thing :smiley_cat:

PR #724, opened 2018-11-04
Maintainance Tasks

LeFnord:
- updates dependencies
- looses rubocop dependency to minor
- adds 2.6.0-preview2 to travis
- adds grape 1.1.0 to travis
- skips testing of grape < 0.19.2
  -> only have last three minors and HEAD


Gemfile
```diff
@@ -26,7 +26,12 @@ group :development, :test do
   gem 'rake'
   gem 'rdoc'
   gem 'rspec', '~> 3.0'
-  gem 'rubocop', '~> 0.60.0', require: false
+  # TODO: change back after 2.6.0 release and updated rubocop version
```
dblock: I wouldn't do this but 🤷‍♂️ :)

LeFnord: … or adding it to `allow_failures`

dblock: I meant I lock my rubocops. I like my builds to not fail in the future :)

PR #729, opened 2018-12-03
Allow empty security array for endpoints

fotos:
Fixes a bug where empty arrays for endpoints (operations) are excluded from the generated schema.

The [OpenAPI 2.0 spec for `security`](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#swagger-object):

> A declaration of which security schemes are applied for the API as a whole. The list of values describes alternative security schemes that can be used (that is, there is a logical OR between the security requirements). **Individual operations can override this definition.**

This is useful in the following scenario:

You define a top level `security` requirement (e.g. `OAuth2`) that applies to the whole API and then selectively whitelist endpoints that are public.

For example:

```
securityDefinitions:
  oauth:
    type: oauth2
    flow: password
    tokenUrl: 'http://127.0.0.1/oauth/token'
    scopes:
      read: Grants read access
      write: Grants write access
security:
  - oauth:
      - read

paths:
  /users:
    post:
      summary: Create User
      security: []
```

Which renders like this in `Swagger UI`:

public

*Note the absence of 🔓 for `/users`.*

There are a couple of more places where `#blank?` is used in `Grape::Endpoint`. Namely in `#swagger_object`, `#info_object`, `#license_object`, and `#contact_object`. I didn't change those since, for example, the `#license_object` can't be blank ([needs at least a `name`](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#license-object)).

This change will produce empty arrays in `#method_object`, for keys such as `parameters` or `tags`. The generated schema is still valid since an empty `parameters` or `tags` is allowed. Especially for parameters it's allowed to override / amend (if defined) but not to remove. The empty array is successfully ignored by `Swagger UI`. This PR will make the generated schema a bit more verbose yet, still, valid. Another alternative is to call [`Object#presence`](https://api.rubyonrails.org/v5.2/classes/Object.html#method-i-presence) in `#params_object` and `#tag_object`.

In the last commit I remove the default empty `description` to bring the generated schema (and specs) closer to what they were originally (before this PR).

☝️ Let me know whether I should include / make those changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 #### Fixes
 
 * Your contribution here.
+* [#729](https://github.com/ruby-grape/grape-swagger/pull/729): Allow empty security array for endpoints - [@fotos](https://github.com/fotos).
 
 ### 0.32.0 (November 26, 2018)
 
--- a/lib/grape-swagger/endpoint.rb
+++ b/lib/grape-swagger/endpoint.rb
@@ -125,7 +125,7 @@ module Grape
       method[:tags]        = route.options.fetch(:tags, tag_object(route, path))
       method[:operationId] = GrapeSwagger::DocMethods::OperationId.build(route, path)
       method[:deprecated] = deprecated_object(route)
-      method.delete_if { |_, value| value.blank? }
+      method.delete_if { |_, value| value.nil? }
 
       [route.request_method.downcase.to_sym, method]
     end
--- a/lib/grape-swagger/endpoint.rb
+++ b/lib/grape-swagger/endpoint.rb
@@ -149,7 +149,6 @@ module Grape
     def description_object(route)
       description = route.description if route.description.present?
       description = route.options[:detail] if route.options.key?(:detail)
-      description ||= ''
 
       description
     end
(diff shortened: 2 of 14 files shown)


fotos:
@dblock and/or @LeFnord, when you get the chance, I would appreciate some feedback for this PR.

I'd rather have a fix upstream (here 😄) than having to fork the project for a small change / fix.

_(Given the opportunity, thanks for all the hard work on Grape and the surrounding ecosystem 🎉)_

dblock:
This makes total sense, I have merged this.

dblock:
@fotos You can make another PR for the remaining things with tests, please. Much appreciated, thanks for the good work!

fotos:
@dblock thanks. 🙇

I'll iterate on another PR and trim the unnecessary `parameters` / `tags`.

Is it possible to get `0.32.1` out? 🙏

dblock:
Happy to make a release soon if @LeFnord is not around, let's see that other PR :)

fotos:
@dblock here you go: https://github.com/ruby-grape/grape-swagger/pull/731 😃

LeFnord:
I'm here, but need a bit to have a look on all the new stuff 😉

PR #733, opened 2018-12-11
Add ability to override data type in swagger when using Custom Types

shanempope:
```ruby
def self.custom_type_data_type_override
  'Integer'
end
```
on your [custom type](https://github.com/ruby-grape/grape#custom-types-and-coercions) will allow you to override the datatype. This would allow you to define it on the custom type once and not have to add documentation hash for every require that uses the custom type.

### DIFF
--- a/lib/grape-swagger/doc_methods/data_type.rb
+++ b/lib/grape-swagger/doc_methods/data_type.rb
@@ -60,8 +60,8 @@ module GrapeSwagger
                     rescue NameError
                       nil
                     end
-            if klass.respond_to?(:custom_type_data_type_override)
-              klass.custom_type_data_type_override.camelize
+            if klass.respond_to?(:data_type)
+              klass.data_type.camelize
             else
               model.to_s.split('::').last
             end
--- a/lib/grape-swagger/doc_methods/data_type.rb
+++ b/lib/grape-swagger/doc_methods/data_type.rb
@@ -55,7 +55,16 @@ module GrapeSwagger
           elsif model.respond_to?(:name)
             model.name.demodulize.camelize
           else
-            model.to_s.split('::').last
+            klass = begin
+                      Object.const_get(model, false)
+                    rescue NameError
+                      nil
+                    end
+            if klass.respond_to?(:custom_type_data_type_override)
+              klass.custom_type_data_type_override.camelize
+            else
+              model.to_s.split('::').last
+            end
           end
         end
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 #### Features
 
-* Your contribution here.
+* [#733](https://github.com/ruby-grape/grape-swagger/pull/733): Override a Custom Type data type with `:custom_type_data_type_override` - [@shanempope](https://github.com/shanempope).
 
 #### Fixes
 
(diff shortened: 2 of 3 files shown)


shanempope:
Something like this would be useful, but for now I'm adding a documentation hash to every `require`.

lib/grape-swagger/doc_methods/data_type.rb
```diff
@@ -55,7 +55,16 @@ def parse_entity_name(model)
           elsif model.respond_to?(:name)
             model.name.demodulize.camelize
           else
-            model.to_s.split('::').last
+            klass = begin
```
dblock: This seems unnecessarily complicated. Why doesn't `model` just implement `.data_type` for example?

dblock:
Could you please provide an example of how this would be used and where it would be useful?

lib/grape-swagger/doc_methods/data_type.rb
shanempope: when using a custom types like the SecureUri example here: https://github.com/ruby-grape/grape#custom-types-and-coercions
Model comes into this function as a string and not an object. This is probably me missing something critical if this seems silly.

shanempope:
@dblock Hey! This is quite possibly me not understanding the framework well enough.
I have a Grape `Custom Type`
```ruby
class Api::Types::PositiveInt32
  MAX_INTEGER_32 = 2_147_483_647

  def self.parse(string)
    begin
      int = Integer(string, 10)
    rescue ArgumentError
      raise ArgumentError.new("Value not a number: #{string}")
    end
    raise ArgumentError.new("Value out of bounds: #{int}") unless int.between?(0, MAX_INTEGER_32)
    int
  end

  def self.parsed?(value)
    value.is_a?(Integer) && value.between?(0, MAX_INTEGER_32)
  end
end
```

and right now if I use that custom type in a `Grape::Api` like this:
```
params do
  requires :id, type: Api::Types::PositiveInt32
end
```
The grape swagger documentation shows the data_type column as `undefined`.
The method I modified in this PR ends up using the final case in `parse_entity_name`, `model.to_s.split('::').last` where it splits and returns `PositiveInt32` which then shows up as `undefined` in the html page of the swagger docs. The model that that method gets is `Api::Types::PositiveInt32` as a String not as an object.

lib/grape-swagger/doc_methods/data_type.rb
shanempope: In line 37 above, in `parse_multi_type(raw_data_type)` they basically do the same thing                       `Object.const_get(model, false)` and `rescue NameError`

shanempope:
@lefnord good point. that method already returns non-spec values though. The final else case just returns the name of the class of the custom type which doesn't match spec. That validity check seems to happen somewhere else after this.

shanempope:
If yall have a better way to support `data type` override for Grape Custom Types, I'm all ears. Going to close this for now because I think it's sloppy.

dblock:
> @shanempope …
> how do you ensure, that only valid datatypes, in the maening of [OAPI spec v2.0#Data Types](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#data-types) will be generated?

This can be left as an exercise :)

dblock:
I am a little uncomfortable with `Object.const_get(model, false)` in this code, otherwise it felt ok to me. 

I have no strong opinions on whether this should be closed or left alone, deferring entirely to @LeFnord.

Issue #730, opened 2018-12-05
swagger is not detecting mounted rack app


bhaskarshankarling:
**Code**
```ruby
require "bundler/setup"
Bundler.require

require 'grape-swagger/entity'

module API
  module V1
    class Welcome < Grape::API
      desc 'Greets user' do
        detail 'This is the root api and it will greet user on accessing'
      end
      get "/" do
        { 
          data: [
            { message: "Welcome to notes app" }
          ]
        }
      end
    end
  end
end

module API
  module V1
    class Base < Grape::API
      version :v1, using: :path

      mount Welcome
    end
  end
end

module API
  class Base < Grape::API
    format :json
    prefix :api

    mount V1::Base

    add_swagger_documentation hide_documentation_path: true,
                              version: "V1",
                              info: {
                                title: 'User notes app',
                                description: 'Demo app for user notes'
                              }
  end
end
```
**Gemfile**
```ruby
# frozen_string_literal: true

source "https://rubygems.org"

ruby '2.5.3'

gem 'grape', '~> 1.2', '>= 1.2.1'

gem 'grape-swagger-entity'

gem 'rake'

gem 'rack-cors'

gem 'pry'
```
**Gemfile.lock**
```
GEM
  remote: https://rubygems.org/
  specs:
    activesupport (5.2.1.1)
      concurrent-ruby (~> 1.0, >= 1.0.2)
      i18n (>= 0.7, < 2)
      minitest (~> 5.1)
      tzinfo (~> 1.1)
    axiom-types (0.1.1)
      descendants_tracker (~> 0.0.4)
      ice_nine (~> 0.11.0)
      thread_safe (~> 0.3, >= 0.3.1)
    builder (3.2.3)
    coderay (1.1.2)
    coercible (1.0.0)
      descendants_tracker (~> 0.0.1)
    concurrent-ruby (1.1.3)
    descendants_tracker (0.0.4)
      thread_safe (~> 0.3, >= 0.3.1)
    equalizer (0.0.11)
    grape (1.2.1)
      activesupport
      builder
      mustermann-grape (~> 1.0.0)
      rack (>= 1.3.0)
      rack-accept
      virtus (>= 1.0.0)
    grape-entity (0.7.1)
      activesupport (>= 4.0)
      multi_json (>= 1.3.2)
    grape-swagger (0.32.0)
      grape (>= 0.16.2)
    grape-swagger-entity (0.3.1)
      grape-entity (>= 0.5.0)
[...]

PR #735, opened 2018-12-19
Added the spec under spec/issues for the issue id 730

bhaskarshankarling:
Added the spec to check the issue #730 related to swagger doc not getting generated when one grape API app is mounted in another grape API app.

### DIFF
--- a/spec/issues/730_swagger_is_not_detecting_mounted_rack_app.rb
+++ b/spec/issues/730_swagger_is_not_detecting_mounted_rack_app.rb
@@ -1,0 +1,77 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+require 'grape-swagger/entity'
+
+module API
+  module V1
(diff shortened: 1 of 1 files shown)


LeFnord:
mmh … shouldn't it be fail?

bhaskarshankarling:
> mmh … shouldn't it be fail?

Yes, I am also wondering why it passed. Since I am new to CI and everything, I thought it is showing green only for removing offences pointed out by rubocop. The second spec is still failing at my local.

bhaskarshankarling:
> mmh … shouldn't it be fail?
https://github.com/ruby-grape/grape-swagger/pull/735#issuecomment-448542399. What about this previous comment generated by bot, it says that it appreciates the failing spec, that means the spec is failing.

bhaskarshankarling:
> mmh … shouldn't it be fail?

Please let me know if I have done anything wrong in the specs.

PR #737, opened 2019-01-03
add the swagger_endpoint_guard to both documentation endpoints

urkle:
adds the swagger_endpoint_guard to both documentation endpoints.


LeFnord:
think it only completes/corrects behaviour, so for both grape-swagger endpoints it will be checked if an gurad clause exists … @urkle, please correct me, if I'm wrong

dblock:
I see, it's just a fix. Let's just merge it.

PR #738, opened 2019-01-21
Allow additional properties to be set

bobbytables:
This is in the OAS 2.0 and 3.0 spec and allows to define an object that is free-form for a property using `additionalProperties` on the documentation.

https://swagger.io/specification/v2/#schemaObject
https://swagger.io/docs/specification/data-models/dictionaries/


LeFnord:
thanks @bobbytables … but, please add some spec, to have documentation of this feature 😉

bobbytables:
Sorry I haven't been able to get to this :( We've been trying to launch our product and hopefully I can make some specs for this.

bobbytables:
I just took a look at where I might need to do this, and please don't take this the wrong way, the spec file for `move_params_spec.rb` is pretty confusing and I have no idea where to even begin. Can you please point me at the right file / area to write this? I see things like private methods getting tested and I really dont want to test this there.

bobbytables:
Would it be possible to be pointed in the right direct? I really don't know where to begin to add a spec for this. For whatever it's worth, we've been using this in production since this PR with no issues too.

LeFnord:
a spec is always part of the documentation of the behaviour of a system, saying something is in production isn't enough

starting by a simple API which uses this feature, have a look (for example) [here](https://github.com/ruby-grape/grape-swagger/blob/master/spec/swagger_v2/api_swagger_v2_request_params_fix_spec.rb)

LeFnord:
Hi @bobbytables … will mörge it for now, cause it seems a valuable feature … thanks for contributing

Issue #603, opened 2017-04-14
Swagger UI 3.0


zhangalex:
Hi Guys, 

This is really a great gem. Thank you.

Since swagger ui 3.0 already released, could you please tell me is there any plan on swagger UI v3?

THANKS AGAIN.

dblock:
Nobody is working on this AFAIK. We look forward to you contributing!

LeFnord:
yeap, nobody is official working on it, so please feel free to contribute 

btw … you should reference the [OAPI spec](https://github.com/OAI/OpenAPI-Specification/blob/OpenAPI.next/versions/3.0.md) itself, not the swagger UI, this isn't a official implementation at the moment, see: [implementations](https://github.com/OAI/OpenAPI-Specification/blob/OpenAPI.next/IMPLEMENTATIONS.md)

blakepettersson:
I have a local branch that works with OpenAPI 3. It's a bit rough around the edges, and I'd need some help to try to see what the next steps would be..

LeFnord:
sounds good 😄 … can you open a PR, so others can have a look on it and helping

run27017:
Does not update any more?

mscrivo:
@blakepettersson @LeFnord Really sad to not see this get over the line, you did so much great work on it! Is there a summary or list of what's left to do to get this done in case someone else wants to pick up the rest and run with it?

We're using grape and finding we're running into so many issues with the lack of OpenAPI v3 support, from no discriminated unions, to no param examples, to no bearer auth support in the security spec, etc. Converting from v2 to v3 only gets you so far and leaves a lot to be desired.

blakepettersson:
@mscrivo it's been years since I've worked on anything Ruby-related, let alone this, so I longer have no idea on what is/was missing. Good luck for anyone wanting to further build upon it!

LeFnord:
@mscrivo … actual, there is no such kind like a to do list for open steps

but there are a bit of work on it, see the [OAPI-v3 branch](https://github.com/ruby-grape/grape-swagger/tree/oapi-3)
feel free to start there and contribute

numbata:
I'd recommend waiting a bit — I'm currently working hard on a new approach to integrate OpenAPI v3 into Grape.

numbata:
From my side this is now addressed in #969 - OpenAPI 3.0 and 3.1 support with full backward compatibility.

jamesmacwilliam:
[...]

PR #744, opened 2019-04-02
Openapi 3.0 [DRAFT]

blakepettersson:
This is an attempt to address #603. OpenAPI 3 is quite different from Swagger 2, which leads to a couple of issues with this PR.

* This PR depends on a patched `grape-swagger-entity` (which is not in this PR)
* `bundle exec rake` *DOES NOT* currently work, since there's currently an issue with running the Swagger 2 and OpenAPI 3 tests at the same time...
* Currently the Group Array Params specs as well as the Mounted spec for OpenAPI 3 are not working (that can be addressed though)

In short, given how different OpenAPI 3 is from Swagger 2, it might be worth having a separate project for this, especially since we need to have a version of `grape-swagger-entity` that's works with OpenAPI3 (which implies that the same time that it won't work for Swagger 2).
 
It was a while since I worked on this, so there are for sure some other issues with this code, which I'll try to address once I remember what I've done.


swistaczek:
Great work! 🤞for future work landing in upstream!

LeFnord:
Hi @blakepettersson … many thanks for the great work 😄 

I've created a [oapi-3](https://github.com/ruby-grape/grape-swagger/tree/oapi-3) branch, and re-target it gainst it

LeFnord:
ok … let us mörge it

gkirill:
Guys, sorry, any suggestions how to enable that? By default it seems to be using 2.0

LeFnord:
use it in your Gemfile by specifying the repo and branch → https://bundler.io/guides/git.html

eistomin:
@LeFnord , can i help you with master tree merging(ruby-grape:oapi-3 -> master or vice versa)?

run27017:
It stills not merged into master?

j-monteiro:
Are there any plans to merge this into master?

EugeneIstomin:
Hi all!
Can we update & merge this branch? It would be like a 2-3 devs collaboration in upditing, merging, testing.

PR #746, opened 2019-06-03
Fix path with optional format

fnordfish:
my issue was raising `undefined method `delete!' for nil:NilClass` for my version path "V2(.{format})"


fnordfish:
The build brakes b/c of some Rubocop rule name has changed :(

dblock:
It's because we're locking down rubocop at a major version and there's a change ...

You can run `rubocop -a ; rubocop --auto-gen-config` and it should fix things.

LeFnord:
thanks @fnordfish

PR #748, opened 2019-06-21
Prepare 0.32.2 release

usernam3:
Hi everybody, thank you very much for working on this great gem.
The last release was more than 6 months ago and the master branch already contains a bunch of useful fixes.
So maybe it makes sense to release the new version which contains those fixes?


LeFnord:
thanks … it is planned for this evening (CET)

Issue #751, opened 2019-07-16
Deeply nested objects


braktar:
The model we use for our API is strongly nested, we are working in Vehicle Routing Optimization and define a complex model in a single request. There are multiple high level objects such as
Services (missions to perform within a route) and Points (coordinates or matrix reference)

Within those objects are others objects
an Activity for Services which have a mission to perform or Activities, if we have alternatives available to perform a delivery.

Points have a single object below, which is the location (lat/lon)

The model is correctly validated with grape and works like a charm at this purpose.
But the generation of the documentation through grape-swagger generate some incorrect types

Location should be an hash, but is displayed as an array of hash.
Skills related to Services are displayed as String and should be array of string
Activity is displayed as an array of hash, and should be a single hash.
Activities are correctly displayed as an array of hash.

braktar:
Please see the related pull request which display the errors.
Has anyone any hints on how I could to solve it or by my side, or by contributing if the bug is confirmed ?

braktar:
By the way, it seems related to the following issues : 
#628 and #666

PR #752, opened 2019-07-16
Deeply nested objects

braktar:
Related to: https://github.com/ruby-grape/grape-swagger/issues/751


braktar:
A bit of progress on my side. My nested objects have the correct type/format which is encouraging.

braktar:
I've also added a nested object with Array[Array[String]]

braktar:
So, maybe a bit of explanation.

MoveParams was following the formData structure in order to build the definition, losing a lot of information useful for the body format.

So instead of it, the definition is now build with the body format (without losing information) and is then modified only if the formData is required.

LeFnord:
thanks @braktar … good job

PR #756, opened 2019-08-28
Fix reference creation when two types are provided

bikolya:
Previously the type from param definition would be used to documentation even if we try to override it in the `documentation` hash.
It caused Swagger UI to render, for example, "Unknown Type: Color" instead of using the provided entity.

It terms of code, before this change we would return something like this:
`{ color: { type: 'Color' }`
instead of:
`{ color: { '$ref' => 'ColorEntity' }`
if `Color` is not inherited from `Grape::Entity`

Having different types can be useful when two different classes are used for parsing and presenting an attribute.
E.g. the input format for Base64 File can be represented as a Hash (content, content_type and filename), whereas the type after parsing can be `ActionDispatch::Http::UploadedFile`.

In the current situation the solution to this problem is to put coercion methods to the same entity and specify `type` only for the param without using `documentation`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 #### Fixes
 
-* Your contribution here.
+* [#756](https://github.com/ruby-grape/grape-swagger/pull/756): Fix reference creation when custom type for documentation is provided - [@bikolya](https://github.com/bikolya).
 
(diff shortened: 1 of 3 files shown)


LeFnord:
thanks @bikolya

PR #757, opened 2019-09-02
Fix array_use_braces for body params

bikolya:
Previously this is what was generated:
`"attr":{"type":"array","items":{"type":"object","properties":{"":{"type":"string"}}}}`
instead of
`"attr":{"type":"array","items":{"type":"string"}}`
in case this attribute was nested in a Hash attribute with "body" in `param_type`

Definition: 
```ruby
requires :param_1, type: Hash, documentation: { param_type: 'body' } do
  optional :param_2, type: Integer
  optional :param_3, type: Array[String]
end
```

Here is how it looked like it Swagger UI: 
image

instead of 
image

### DIFF
--- a/lib/grape-swagger/endpoint/params_parser.rb
+++ b/lib/grape-swagger/endpoint/params_parser.rb
@@ -15,17 +15,14 @@ module GrapeSwagger
       end
 
       def parse_request_params
-        array_keys = []
         public_params.each_with_object({}) do |(name, options), memo|
           name = name.to_s
           param_type = options[:type]
           param_type = param_type.to_s unless param_type.nil?
 
           if param_type_is_array?(param_type)
-            array_keys << name
             options[:is_array] = true
-
-            name += '[]' if array_use_braces?(options)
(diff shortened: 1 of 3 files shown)


lib/grape-swagger/endpoint/params_parser.rb
```diff
@@ -34,8 +31,8 @@ def parse_request_params
 
       private
 
-      def array_use_braces?(options)
-        settings[:array_use_braces] && !(options[:documentation] && options[:documentation][:param_type] == 'body')
+      def array_use_braces?
+        @array_use_braces ||= settings[:array_use_braces] && !includes_body_param?
```
LeFnord: where are the settings come from, ist it configurable?
if yes, then it should be described in the README

bikolya: It's described here: https://github.com/ruby-grape/grape-swagger#array_use_braces

LeFnord:
thanks @bikolya …

PR #758, opened 2019-09-04
Handle cases where a route's prefix is a nested URL

SimonKaluza:
Nested prefixes (any prefix containing a '/' character) prior to this commit would incorrectly get tagged as the first segment of the URL prefix.

E.g. a route like "api/beta/v1/patients" would incorrectly get tagged as "api" instead of "patients"

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 #### Fixes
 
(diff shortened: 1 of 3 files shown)


LeFnord:
mmh for me the changes are reasonable, but please adept the specs and/or add one which documents the behaviour

SimonKaluza:
Added a RSpec example with tags that breaks without the adjustments made to endpoint.rb to support nested prefix and rebased off of the current master (hence the force-pushes)

LeFnord:
thanks @SimonKaluza

PR #760, opened 2019-09-30
GrapeSwagger::DocMethods::Extensions breaks the contract of Kernel#method

timgaleckas:
This causes code inspection tools and development workflows to break
and may cause developers to choose not to use the gem.

One such tool that breaks is https://sorbet.org/


LeFnord:
please add a spec as documentation waht this change implies

maybe changing the the PR description into a more understandable one, so one without the knowledge what `sorbet` means (besides something to eat) get an idea, thanks

LeFnord:
ah … thanks @timgaleckas

timgaleckas:
Thank you.

PR #761, opened 2019-10-04
Add an option to configure root element for responses 

bikolya:
It would be nicer to have an option similar to `is_array` in the desc block instead of using `route_setting`, but apparently it requires a change in the grape repository similar to this one: https://github.com/ruby-grape/grape/pull/1791

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,9 +3,11 @@
 #### Features
 
 * Your contribution here.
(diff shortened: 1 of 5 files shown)


LeFnord:
thanks @bikolya … and sorry for the late reply … after running green I'll mörge it

LeFnord:
sorry overseen the last line, did you created a PR for grape to support this option?  
If not how we should handle it?

Issue #750, opened 2019-07-03
Version 0.33 requires ruby 2.4 and above


sondnm:
Firstly thanks a lot for the contributions of this awesome gem.

Recently I was trying to upgrade this gem to version `0.33` but got this error while running `bundle update`. My ruby version is `2.3.3`.

```
Bundler could not find compatible versions for gem "ruby":
  In Gemfile:
    ruby

    grape-swagger (= 0.33) was resolved to 0.33.0, which depends on
      ruby (>= 2.4)
```

As I understand, from version `0.33`, grape-swagger will not support ruby below 2.4. Since this is not mentioned in the CHANGELOG, I would like to know whether you actually drop support for ruby below 2.4 or not.

Thank you very much.

myxoh:
This doesn't seem to be correct. It seems this was changed in a `prepare to release` commit, rather than on a PR:
https://github.com/ruby-grape/grape-swagger/commit/7237ef3525f907b82b1b4be8d81269d3602af7d0
@LeFnord  did you do so in purpose? If so - can we ammend the changelog? Else - can we rever this?

LeFnord:
yeap, offcial support for ruby < 2.4 is dropped as it is common for most gems to only support last three versions … a look into travis ci shows the supported versions

@myxoh … yeap I dropped it … if you want, add something to the changelog, but please don't revert it, there was a bit more on changes as you mentioned

@sondnm … maybe it is a good point to update your used ruby version, the actual one is 2.6 and 2.7 is in the pipeline

sondnm:
Thanks a lot for the confirmation.
I am having trouble updating the ruby version although I know it's no longer supported 😆

I created a PR to update the changelog https://github.com/ruby-grape/grape-swagger/pull/763 though.

PR #763, opened 2019-10-20
Add ruby version changes to CHANGELOG

sondnm:
Since version 0.33, we dropped support for ruby 2.3 and below but hasn't updated the changelog. Adding a docs to notify users about this change and avoid confusion similar to https://github.com/ruby-grape/grape-swagger/issues/750


LeFnord:
thanks … @sondnm … and yeap same play for 2.4 in Januar, after 2.7 is out

PR #764, opened 2019-10-30
Replace downcase with underscore for multi-word roots

bikolya:
It fixes a problem introduced in #761 
When `downcase` was used, `apierror` instead of `api_error` was generated from ApiError entity.


LeFnord:
thanks … @bikolya, welcome 😄

 and sorry for late replay

LeFnord:
a bit curios … will have a look on it tonight

dblock:
The CHANGELOG complaint is right, you're missing a colon after the PR :)

LeFnord:
ah ok … thanks

PR #768, opened 2020-01-03
Uses ruby 2.7

LeFnord:


LeFnord:
[correct]: it will be the last one, which supports grape < 1.3.0, cause of breaking changes → replacing virtus by dry-type

PR #773, opened 2020-01-20
Freeze Rack version to 2.0.8

LeFnord:
- corrects version matrix


urkle:
@LeFnord why are you freezing rack to a very specific version? Wouldn't having it as a minimum required be better? As with this if I have another gem in my app that needs, say, 2.0.9, I will be unable to have that and grape at the same time since this is exposed in the gemspec.

mobilutz:
@urkle @LeFnord I am also wandering why rack is being frozen to a specific version?

grape-swagger.gemspec
```diff
@@ -15,6 +15,7 @@ Gem::Specification.new do |s|
 
   s.required_ruby_version = '>= 2.4'
   s.add_runtime_dependency 'grape', '>= 0.16.2', '< 1.3.0'
+  s.add_runtime_dependency 'rack', '2.0.8'
```
dblock: This locks dependency to a specific version of Rack. Definitely don't want to do that as it will render grape-swagger incompatible with most apps. 

Grape 1.3 is compatible with rack 2.1, there's a matrix where you can make it all work. LMK if you need help.

LeFnord: Yeap I know, the problem here isn’t grape, it is grape-swagger and I’m on it.

Sorry for circumstances, if you need rack > 2.0.x use version 0.34.1, there rack isn’t frozen.

swistaczek:
Hey, why this version has been frozen?

LeFnord:
please read the comment above

PR #775, opened 2020-01-23
allow token_owner to be passed to the hidden proc for parameters as well

urkle:
This allows the same functionality that is offered by the route-level hidden procs.


LeFnord:
thanks @urkle … will be mörged

Issue #657, opened 2017-12-12
Using an entity to represent a post body in params


eropple:
Hey there! Thanks so much for all the hard work that goes into `grape-swagger` and its related projects; they're a huge help.

One thing that I've been running into, and it's something of a cross-cutting concern so I wasn't quite sure whether to put it here or in `grape-swagger-entity`, is a common problem I'm running into. In other frameworks I've used, you can specify a body's parameter set as an entity or an object. For example, I might have `LoginRequest`, which specifies `username`, `plaintext`, that sort of thing. However, `grape-swagger` does not make it obvious as to how I would do this and I get a lot of weird error messages (code deep in `grape-swagger` that gets nils it's not expecting, etc.) when I try to just say "the body parameters for this are just _this_ entity right here". 

I know Swagger can describe it, because I've written Swagger documents by hand that do so. I just can't figure out how to make `grape-swagger` do it, or if the library can at all? Maybe this is totally doable and I'm just goofy, and if so I would love a pointer! But the best I've found is to have a single parameter, conventionally `request`, and dig into `params[:request]` for it. Which isn't _bad_, but the experience of it is kind of weird when I use a client library where I'm passing a named `request` parameter instead of just `username` and `plaintext`. It feels more clunky than the rest of Grape and `grape-swagger` so I'm wondering either if I'm missing something or if this is as-designed.

Thanks again!

eropple:
Bump; it's been about a month and some guidance would be really appreciated.

Bugagazavr:
@eropple Hello, can you provide some examples of code or spec for this feature?

eropple:
I'd like to generate an option that looks like this (from the canonical `petstore-expanded.yaml`):

```yaml
    post:
      description: Creates a new pet in the store.  Duplicates are allowed
      operationId: addPet
      parameters:
        - name: pet
          in: body
          description: Pet to add to the store
          required: true
          schema:
            $ref: '#/definitions/NewPet'
```

I can't for the life of me figure out how to generate this.

bobbytables:
I'm in the same boat. All of the swagger doc is generated with the params set to `in: "formData"` when it's actually in the JSON payload.

LeFnord:
[...]

PR #778, opened 2020-02-21
Add params model in route options

PhilippeChab:
See https://github.com/ruby-grape/grape-swagger/issues/657#issuecomment-581653945 for the reason behind this PR.

This solution is not the cleanest but it actually give us a way to have a decent model name in the documentation. It can be temporary.

I didn't read the source code all over, but I don't understand why the params key doesn't have the same behavior as the success one. I think it could be a permanent solution.

### DIFF
--- a/lib/grape-swagger/doc_methods/move_params.rb
+++ b/lib/grape-swagger/doc_methods/move_params.rb
@@ -38,8 +38,9 @@ module GrapeSwagger
         end
 
         def parent_definition_of_params(params, path, route)
-          definition_name = OperationId.manipulate(parse_model(path))
-          referenced_definition = build_definition(definition_name, params, route.request_method.downcase)
+          has_params_model_name = route.options[:params_model_name].present?
+          definition_name = has_params_model_name ? route.options[:params_model_name] : OperationId.manipulate(parse_model(path))
+          referenced_definition = build_definition(definition_name, params, has_params_model_name ? nil : route.request_method.downcase)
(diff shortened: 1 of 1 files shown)


PhilippeChab:
I've experimented more with grape, grape-swagger and grape-entity and I just don't understand the way it's designed at the moment. 

Why would you want to add a ```params``` key in your ```desc``` block like this:
```
desc: "Some endpoint",
    params: API::V1::Entities::SomeEntity,
    ...
get ":id" do
...
```
since you can already add documentation to your params with grape this way:
```
params do
    requires :some_param, type: String, documentation: { desc: "Some description", in: "body" }
    ...
get ":id" do
...
```
Because encapsulating the params into an entity is a cleaner way to do this? Yes, I agree. But the problem is, the params in the ```desc: ...``` block are not validated, so you absolutely need to add the route params in the ```params``` block. What is the point to document the params in both block then? I don't see any reason, which would mean the ```params``` key in the ```desc``` block is useless. If there is something I am missing, please tell.

That being said, I still think entities are a really nice facade to endpoint responses and I don't understand why they can't be use directly in grape ```params``` block for endpoint params. Why can't we do something like this:
```
des "...",
    ....
params API::V1::Entities::SomeEntity
get ":id" do
    ...
```
This would make an entity really great, since it could provide validation, documentation and encapsulation all at once.

I'm closing this PR since I don't see any reason to integrate more support to the ```params``` key.

PR #782, opened 2020-03-23
Allow passing class name as a string for oapi rake task initializer

misdoro:
It may be possible that API classes are not yet loaded when the oapi rake tasks are initialized in Rakefile.

Add a possibility to pass class name as a string to not mess with class loader configuration.


LeFnord:
very cool … thanks @misdoro

PR #786, opened 2020-04-18
Use full entity name as a default (fixes #779)

### LINKED ISSUE #779: Incomplete entity name for swagger documentation
I've met a problem with swagger documentation because of this method, that parses model name and leaves the last exact entity name (if it is nested). It might be a problem for the case I am going to describe.

https://github.com/ruby-grape/grape-swagger/blob/7237ef3525f907b82b1b4be8d81269d3602af7d0/lib/grape-swagger/endpoint.rb#L335-L337

For my project I use this versioning approach:

```ruby
module API
  module SomeModel
    class V1 < Grape::Entity
      ...
    end
    
    class V2 < Grape::Entity
      ...
    end
  end
end
```

So, for that case parsing of a  model to model name returns just `"V1"` or `"V2"`. That is why all endpoints that end with `V1` will represent the last parsed endpoint.

Could the parsed model name be widened to all the described entity name?

mrexox:
In this PR I just wanted to end with this:

```
- model.to_s.split('::').last
+ model.to_s
```

But I met a problem that it receives Class type in here
https://github.com/ruby-grape/grape-swagger/blob/8abddd5f2ce5fe29238990b136625f34371b0be0/lib/grape-swagger/endpoint.rb#L354
But String type in the another place. So, splitting by `::` and taking the last word actually returns the same as `name` method of the class. This is fine, but...

For example, if we use such constructions, we end up with just `V1` and `V2` from the `A::B` module as it is parsed first:

```ruby
module A
  module B
    class V1 < Grape::Entity
     ...

    class V2 < Grape::Entity
     ...

module A
  module C
    class V1 < Grape::Entity
      ...

    class V2 < Grape::Entity
      ...
```

This PR fixes this [issue](https://github.com/ruby-grape/grape-swagger/issues/779).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,8 @@
 * [#785](https://github.com/ruby-grape/grape-swagger/pull/785): Add extensions for params - [@MaximeRDY](https://github.com/MaximeRDY).
 * [#782](https://github.com/ruby-grape/grape-swagger/pull/782): Allow passing class name as string for rake task initializer - [@misdoro](https://github.com/misdoro).
 
-
 #### Fixes
+* [#786](https://github.com/ruby-grape/grape-swagger/pull/786): Use full entity name as a default - [@mrexox](https://github.com/mrexox)
 
 ### 1.0.0 (February 10, 2020)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,9 +5,9 @@
 * Your contribution here.
 * [#785](https://github.com/ruby-grape/grape-swagger/pull/785): Add extensions for params - [@MaximeRDY](https://github.com/MaximeRDY).
 * [#782](https://github.com/ruby-grape/grape-swagger/pull/782): Allow passing class name as string for rake task initializer - [@misdoro](https://github.com/misdoro).
+* [#786](https://github.com/ruby-grape/grape-swagger/pull/786): Use full entity name as a default - [@mrexox](https://github.com/mrexox)
 
 #### Fixes
-* [#786](https://github.com/ruby-grape/grape-swagger/pull/786): Use full entity name as a default - [@mrexox](https://github.com/mrexox)
 
 ### 1.0.0 (February 10, 2020)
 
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -6,8 +6,8 @@
 * [#785](https://github.com/ruby-grape/grape-swagger/pull/785): Add extensions for params - [@MaximeRDY](https://github.com/MaximeRDY).
 * [#782](https://github.com/ruby-grape/grape-swagger/pull/782): Allow passing class name as string for rake task initializer - [@misdoro](https://github.com/misdoro).
 
-
 #### Fixes
+* [#786](https://github.com/ruby-grape/grape-swagger/pull/786): Use full entity name as a default - [@mrexox](https://github.com/mrexox)
```
LeFnord, line 10: would say its a feature insead of a fix 😉

spec/issues/430_entity_definitions_spec.rb
```diff
@@ -82,11 +82,11 @@ class NameApi < Grape::API
     JSON.parse(last_response.body)['definitions']
   end
 
-  specify { expect(subject).to include 'Class1' }
-  specify { expect(subject).to include 'Class2' }
+  specify { expect(subject).to include 'TestDefinition::DummyEntities::WithVeryLongName::AnotherGroupingModule::Class1' }
```
LeFnord, line 85: yeah, this will be very long …

CHANGELOG.md
mrexox: Yep! Moved a line 👌

spec/issues/430_entity_definitions_spec.rb
mrexox: But correct 😏

mrexox:
@LeFnord , added an upgrade entry. Does it look fine?

spec/issues/430_entity_definitions_spec.rb
LeFnord: yeap …

LeFnord:
will release it tonight

PR #793, opened 2020-06-05
Features/inheritance and discriminator

MaximeRDY:
Hi,

I would like to add inheritance with discriminator like that: 
```
 class Pet < Grape::Entity
          expose :type, documentation: {
            type: 'string',
            is_discriminator: true,
            required: true
          }
          expose :name, documentation: {
            type: 'string',
            required: true
          }
        end

        class Cat < Pet
          expose :huntingSkill, documentation: {
            type: 'string',
            description: 'The measured skill for hunting',
            default: 'lazy',
            values: %w[
              clueless
              lazy
              adventurous
              aggressive
            ]
          }
        end
```
Need this PR: https://github.com/ruby-grape/grape-swagger-entity/pull/50
What do you think about it ?


MaximeRDY:
@LeFnord Any news about that ? :-)

LeFnord:
Hi @MaximeRDY … can you describe for what it is needed/can be used?   
sorry didn't see it, how does it match the specification?

spec/issues/inheritance_and_discriminator_spec.rb
```diff
@@ -0,0 +1,56 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
```
LeFnord, line 3: this isn't an issue, so please move it under `spec/swagger_v2` … thanks

spec/spec_helper.rb
```diff
@@ -30,6 +30,7 @@
 
 require 'rack'
 require 'rack/test'
+require 'pry'
```
LeFnord, line 33: ??? please remove it … running the specs with `bundle exec …` should do it

spec/issues/inheritance_and_discriminator_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+
+    subject do
+      get '/swagger_doc'
+      JSON.parse(last_response.body)['definitions']
+    end
+
+    specify do
+      subject['InheritanceTest::Entities::Pet'].key?('discriminator')
+      subject['InheritanceTest::Entities::Pet']['discriminator'] = 'type'
+      subject['InheritanceTest::Entities::Cat'].key?('allOf')
```
LeFnord, line 53: a ok, now I see it …

Bugagazavr:
Hello, as I know `allOf`, `oneOf`, `anyOf`, `discriminator` it's a part of OAS 3 specification. Maybe the target branch should be oapi-3?

LeFnord:
@Bugagazavr … no not really … I though the same and checked it, but it exists also in 2.0 → [schema-object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#schema-object)

MaximeRDY:
@LeFnord

LeFnord:
@Bugagazavr … can you have a look on https://github.com/ruby-grape/grape-swagger-entity/pull/50, thanks

lib/grape-swagger/doc_methods/build_model_definition.rb
```diff
@@ -17,6 +17,57 @@ def build(model, properties, required)
           definition
         end
 
+        def parse_params_from_model(parsed_response, model, model_name)
```
Bugagazavr, line 20: One of my strong doubts is related to this part of the code. Instead of returning a tuple with a definition and list of attributes, we start to return just a `Hash`. Of course, this because we previously return only the `properties` field, and now we need to return full definition as is.

In my opinion, `grape-swagger-entity` should return the final definition without an addition wrapping ( like we do this in the `build` method ). But I do not insist if it ok for @LeFnord.

MaximeRDY: But if we do that, it result a big breaking change for all parser no ?

Bugagazavr: Right. In my mind, we can declare a wrapper class with a method `to_definition` on the `grape-swagger` side. Parsers will use this class or can use own inherited from this class to declare more complex logic.

For the backward compatibility, we can check is the result of the parser is an array or wrapper object.

MaximeRDY: What do you think @LeFnord ?

MaximeRDY:
Hi @LeFnord  any news ? :-)

lib/grape-swagger/doc_methods/build_model_definition.rb
LeFnord: Moinsen … and sorry for latency …

for this PR I'm fine with the actual solution, but we should change the behaviour as @Bugagazavr mentioned above in another PR to make things clear as a breaking change

Bugagazavr: @LeFnord let me know when you release this feature. I need to release `grape-swagger-entity` with dependency lock to the latest `grape-swagger`.

LeFnord: @Bugagazavr … I can not release it, cause `grape-swagger-entity` references a git branch 

maybe you releasing `grape-swagger-entity` and set the grape-swagger dep to 1.4.0, after that I can change it here, mörge it and make a new release 1.4.0

what di you think?

Bugagazavr: Done https://rubygems.org/gems/grape-swagger-entity

LeFnord:
Hi @MaximeRDY … please have a look at Travis, there are some failures relating to grape-swagger-entity

MaximeRDY:
> Hi @MaximeRDY … please have a look at Travis, there are some failures relating to grape-swagger-entity

Hi @LeFnord Ok I will look on it

LeFnord:
it seems fixed in grape-swagger-entity → see: https://github.com/ruby-grape/grape-swagger-entity/pull/52

MaximeRDY:
> it seems fixed in grape-swagger-entity → see: [ruby-grape/grape-swagger-entity#52](https://github.com/ruby-grape/grape-swagger-entity/pull/52)

Yes I saw that. Its why I've added a commit with no change to force the test :-) 
Otherwise I've change the rubocop and add `Metrics/CyclomaticComplexity` to 14 instead of 13.

LeFnord:
now it should run … after all is green, I mörge it … thanks @MaximeRDY

Issue #659, opened 2018-01-11
Inherited `entity_name` is not possible


jweir:
What is the reasoning for using `methods(false)` to detect if a model has an `entity_name` method?

https://github.com/ruby-grape/grape-swagger/blob/d6bed9523344115e7786e8161de14af93aa067b9/lib/grape-swagger/doc_methods/data_type.rb#L51

In my case we have several Entity classes of the same type, but in different domains. Eastern::Market, Western::Market, etc.  These all inherit from a base Grape::Entity class with a custom entity_name method to output EasternMarket, or WesternMarket, etc.

After making an upgrade from an older version of grape-swagger this breaks.

LeFnord:
closewd via #794

PR #794, opened 2020-06-09
Allow entity_name to be called on parent classes

urkle:
I am unsure why this limitation was done and issue #659 posses the same question.

This PR changes the behavior to allow calling entity_name if a parent class implements it.

My use case for this as follows

```ruby
module Public::V2::Entities
   class Base < Grape::Entity
     # ... other shared logic
     def self.entity_name
       "V2::#{self.to_s.demodulize}"
     end
   end
   
   def MyEntity < Base
     # ....
   end
end
```

And I have ~ 20+ entities all inheriting from base and I don't really want the FULL module tree exposed in the API (in our application it is a little longer than what you see here). And being able to override entity_name once is significantly easier.

### DIFF
--- a/lib/grape-swagger/doc_methods/data_type.rb
+++ b/lib/grape-swagger/doc_methods/data_type.rb
@@ -48,7 +48,7 @@ module GrapeSwagger
         end
 
         def parse_entity_name(model)
-          if model.methods(false).include?(:entity_name)
+          if model.respond_to?(:entity_name)
             model.entity_name
           elsif model.to_s.end_with?('::Entity', '::Entities')
             model.to_s.split('::')[0..-2].join('_')
--- a/spec/lib/data_type_spec.rb
+++ b/spec/lib/data_type_spec.rb
@@ -49,6 +49,18 @@ describe GrapeSwagger::DocMethods::DataType do
     it { is_expected.to eq 'MyInteger' }
   end
 
+  describe 'Types in array with inherited entity_name' do
+    before do
+      stub_const 'EntityBase', Class.new
+      allow(EntityBase).to receive(:entity_name).and_return 'MyInteger'
+      stub_const 'MyEntity', Class.new(EntityBase)
+    end
+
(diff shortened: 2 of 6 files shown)


lib/grape-swagger/doc_methods/data_type.rb
```diff
@@ -48,7 +48,7 @@ def parse_multi_type(raw_data_type)
         end
 
         def parse_entity_name(model)
-          if model.methods(false).include?(:entity_name)
+          if model.respond_to?(:entity_name)
```
LeFnord, line 51: mmh, would `method_defined?` be the choice for it here?

spec/lib/data_type_spec.rb
```diff
@@ -49,6 +49,18 @@
     it { is_expected.to eq 'MyInteger' }
   end
 
+  describe 'Types in array with inherited entity_name' do
```
LeFnord, line 52: mmh, thin [reference_entity_spec.rb](https://github.com/ruby-grape/grape-swagger/blob/master/spec/swagger_v2/reference_entity_spec.rb) will be a better place for it

lib/grape-swagger/doc_methods/data_type.rb
urkle: method_defined? is the old behavior, which only returns true if it is defined on *this* exact class.    (which is the old behavior).

What I'm wanting here is to allow a parent implemented entity_name to be used on child classes.

spec/lib/data_type_spec.rb
urkle: I copied and modified the test that is above this one in this spec. As this is testing the unit of handling calculating the entity.  I can add tests there as well.

LeFnord:
@urkle … please can you have a look on it, want have it in in trhe next release

urkle:
@LeFnord I re-based retested and corrected the rubocop issues

LeFnord:
cool thanks @urkle

PR #795, opened 2020-06-10
properly handle DELETE endpoints

barunio:
Prior to this commit, DELETE endpoints were not being documented
appropriately. Parameters for these endpoints should be documented as
part of the body, not the path.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -11,6 +11,9 @@ AllCops:
 Layout/EmptyLinesAroundArguments:
   Enabled: false
 
+Layout/EmptyLinesAroundAttributeAccessor:
+  Enabled: true
+
 Layout/FirstHashElementIndentation:
   EnforcedStyle: consistent
 
@@ -24,6 +27,12 @@ Layout/SpaceAroundMethodCallOperator:
 
 # Lint stuff
 #
+Lint/DeprecatedOpenSSLConstant:
+  Enabled: true
+
+Lint/MixedRegexpCaptureTypes:
+  Enabled: true
+
 Lint/RaiseException:
   Enabled: true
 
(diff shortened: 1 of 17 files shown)


LeFnord:
Hi @barunio … nice thesis 😉 … 

according to the http specification (RFC 2616) 

"_The DELETE method requests that the origin server delete the resource identified by the Request-URI._" 

is the actual behaviour the correct one, but it can be changed to "let the user choose" which behaviour she wants, also according to the spec 

"_This method MAY be overridden by human intervention (or other means) on the origin server._"

barunio:
@LeFnord I don't think the portion of the spec you're referencing is relevant to the change I'm proposing here. 

The issue I am trying to address is that the json created by `grape-swagger` currently sets `in: query` for parameters passed to `delete` requests, and I contend that these parameters should instead be considered as part of the request body. The latest spec, RFC 7231, [is ambiguous about the idea of request bodies in DELETE requests](https://tools.ietf.org/html/rfc7231#section-4.3.5):

> A payload within a DELETE request message has no defined semantics;
   sending a payload body on a DELETE request might cause some existing
   implementations to reject the request.

In reality, though, all modern browsers support a request body for DELETE requests. And the common practice for modern web applications is to utilize a request body when there are parameters for these requests. So in reality parameters for DELETE requests could _either_ be in the URL or request body, but I believe it is more sensible to assume the latter, not the former.

LeFnord:
the target of my comment wasn't the change itself, it was the wording of your PR description, please think about it

so please implement it that way, that the default is in the path (as part of the URI, as the RFC it describes), but the user can set it to the body

besides that, the browswer ins't the point, the server is it, which have to accept it, from this point, grape don't differ between path, query or body params, that's all params 😉

olivier-thatch:
It is true that RFC 7231 doesn't explicitly forbid DELETE requests from having a body, and Rack itself doesn't care and will read parameters from both query strings and bodies regardless of which HTTP method is used.

That said, conventionally DELETE requests do not have bodies. DELETE requests, like GET requests, are also supposed to be idempotent, and having a body makes that harder IMHO.

@numbata I think this PR should be closed.

PR #796, opened 2020-06-11
Support Grape 1.4.0

thedanielhanke:
dont call me impatient, but it came to my attention that grapes master-branch version is > 1.3.0

Not quite sure about desired version handling, so i guessed the closest to ~> 1.3.0 while supporting 1.4.0 too, is to go >= 1.3.0 < 1.5.0.

### DIFF
--- a/grape-swagger.gemspec
+++ b/grape-swagger.gemspec
@@ -14,7 +14,7 @@ Gem::Specification.new do |s|
   s.license     = 'MIT'
 
   s.required_ruby_version = '>= 2.4'
-  s.add_runtime_dependency 'grape', '~> 1.3.0'
+  s.add_runtime_dependency 'grape', '~> 1.3'
 
   s.files         = `git ls-files`.split("\n")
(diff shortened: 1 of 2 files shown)


grape-swagger.gemspec
```diff
@@ -14,7 +14,7 @@ Gem::Specification.new do |s|
   s.license     = 'MIT'
 
   s.required_ruby_version = '>= 2.4'
-  s.add_runtime_dependency 'grape', '~> 1.3.0'
+  s.add_runtime_dependency 'grape', '>= 1.3.0', '< 1.5.0'
```
LeFnord, line 17: use '~> 1.3' instead, it will do the magic

thedanielhanke: gosh..yep :palm: 
since your comment is longer and more content than the actual change, well, "sorry" for "contributing "  :smile: thanks

LeFnord:
😄 … every commit is a contribution, thanks @thedanielhanke

PR #799, opened 2020-06-30
Points to grape-swagger-entity fix.

LeFnord:
- allows grape head to fail ... for the moment

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -11,6 +11,9 @@ AllCops:
 Layout/EmptyLinesAroundArguments:
   Enabled: false
 
+Layout/EmptyLinesAroundAttributeAccessor:
+  Enabled: true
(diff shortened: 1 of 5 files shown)


LeFnord:
@Bugagazavr … have a look at the travis matrix … and the check

so think I'll mörge this, so other PRs can be finished and make a new release,  
afterwards the fix fur grape head and again a new release … did it make sense?

Bugagazavr:
@LeFnord At first look, it seems ok. But I worry about broken definition specs for `HEAD` builds. 

I will try to look into this tomorrow. If you are sure this not affecting stable versions, it's ok to fix the `HEAD` in the next release.

LeFnord:
@yeap … I'm sure … actual there is much progress in grape development

PR #800, opened 2020-07-01
Prepare release 1.2.0

LeFnord:


LeFnord:
it is allways the same with danger, in every PR it finds an error in the changelog, but nobody knows how an entry should look like and copy&paste doesn't allways solve it … 🤬

PR #801, opened 2020-07-13
Fixes behaviour after grape upgrade to 1.4.0.

LeFnord:


dblock:
@LeFnord, what did Grape 1.4.0 break?

LeFnord:
@dblock … most of the errors depended of this [line](https://github.com/ruby-grape/grape-swagger/pull/801/files#diff-7af39528837333c960830618fb6a52e9R308) …  it seems `declared_params` changed, but this is only my guess

we observed a abd behaviour [here](https://travis-ci.org/github/ruby-grape/grape-swagger/jobs/703524293)

this PR will fix it, but also runs with 1.3.3

dblock:
I bisected it to https://github.com/ruby-grape/grape/pull/2077.

dblock:
Looks like the order of setting internal `:declared_params` has changed, and there was no need to rely on it anymore since Grape < 1.0. I think this is safe to merge.

Issue #578, opened 2017-02-05
Response Model Descriptions are incorrectly set and overwritten.


toshe:
When an endpoint is generated in the swagger doc the `description` field of the response model is set incorrectly:

```ruby
@definitions[response_model][:description] = description_object(route)
```
The description_object(route) method returns the `description` of the route, which can be really long and is different for each route! 

So if you have an endpoint, say, getKittens:

```ruby
desc 'Get all kittens!', {
  nickname: 'getKittens',
  success: Entities::Kitten
  failure: [[401, 'KittenBitesError', Entities::BadKitten]]
  }
get '/kittens' do
   puts "meow"
end
```

Grape Swagger will define the new models Kitten and also BadKitten and will apply the same description in the the definition. Swagger generated doc:

```yml
definitions:
  kitten:
    type: object
    properties:
      fluffy:
        type: boolean
        description: Indicates if the kitten is fluffy or not!
    required:
      - fluffy
    description:  Get all kittens!
```

If you use the `KittenEntity` as a response somewhere else, then the model definition description will get overwritten once more with the description of the route.

Since the description of the route is in no way useful for the model description I suggest we simply replace it with a generic description:

```ruby
@definitions[response_model][:description] = "A #{response_model} model"
```

LeFnord:
@toshe … yeap it makes sense to me, a PR would be nice 😉

LeFnord:
closed via #804 https://github.com/ruby-grape/grape-swagger/pull/804

PR #804, opened 2020-07-28
Don't overwrite model description with the route description.

Bhacaz:
This PR have 3 features around the API response documentation.

1. Removing the assignation of the model description from route description. Some work was already started in this PR https://github.com/ruby-grape/grape-swagger/pull/629.
2. Generate models documentation before the routes.
3. Add the ability to specified no model (or empty body).

### 1. Removing the assignation of the model description from route description.

I face the same problem explain in:
* https://github.com/ruby-grape/grape-swagger/pull/629
* https://github.com/ruby-grape/grape-swagger/issues/578

And I try to continued to resolved it.

Following the PR comments and the bug suggestion, the change in this PR will set a description to the model only if it doesn't have one and with a generic sentence.

### 2. Generate models documentation before the routes.

When I was generating my documentation I observe that sometimes `Grape::Endpoint#response_object` try to find the model definition without succeed with `Grape::Endpoint#expose_params_from_model` simply because the route documentation was generated before the model definition.

### 3. Add the ability to specified no model (or empty body).

The last thing I add was the ability explicitly set no model (or an empty body) in the response, for example in the failures. The problem was that when none is explicitly passed, `Grape::Endpoint#response_object` is really good to found a model from the path and add it to the documentation.

I hope it will be satisfying.

### DIFF
--- a/lib/grape-swagger/endpoint.rb
+++ b/lib/grape-swagger/endpoint.rb
@@ -78,11 +78,11 @@ module Grape
     def path_and_definition_objects(namespace_routes, options)
       @paths = {}
       @definitions = {}
+      add_definitions_from options[:models]
       namespace_routes.each_value do |routes|
         path_item(routes, options)
       end
 
-      add_definitions_from options[:models]
       [@paths, @definitions]
     end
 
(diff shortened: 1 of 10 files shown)


LeFnord:
yeap that does it 😄 … thanks @Bhacaz 

please can you add an entry in [UPGRADING.md](https://github.com/ruby-grape/grape-swagger/blob/master/UPGRADING.md) with title: `Upgrading to >= 1.3.0`, thanks

LeFnord (approved):

Bhacaz:
> please can you add an entry in UPGRADING.md with title: Upgrading to >= 1.3.0

Sure!

Bhacaz:
I needed to refactor [lib/grape-swagger/doc_methods.rb](https://github.com/ruby-grape/grape-swagger/compare/c6cef6a4a9dbe3e2da137ce6fbd9c08824f7503d..fc7b5b126b954c49477e150f59dc71a3ce102820) to satisfy Rubocop (0.90.0)

lib/grape-swagger/endpoint.rb
```diff
@@ -207,19 +207,20 @@ def response_object(route, options)
...
         if memo.key?(200) && route.request_method == 'DELETE' && value[:model].nil?
           memo[204] = memo.delete(200)
           value[:code] = 204
+          next
         end
 
-        next if value[:code] == 204
-        next unless !response_model.start_with?('Swagger_doc') && (@definitions[response_model] || value[:model])
+        # Explicitly request no model with { model: '' }
+        next if value[:model] == ''
```
LeFnord, line 217: mmh … ok … will have a look on it later, maybe it can be get together with no model key given

LeFnord (approved):

LeFnord:
very good job, many thanks @Bhacaz

lib/grape-swagger/endpoint.rb
Bhacaz: @LeFnord It did that because by default if no model key is provided ([line 219](https://github.com/ruby-grape/grape-swagger/pull/804/files#diff-7af39528837333c960830618fb6a52e9R219)) the model use for the response will be the one returned by `GrapeSwagger::DocMethods::PathString#build` called here https://github.com/ruby-grape/grape-swagger/blob/08b8c2fe0518043e4a479453327068ab0285d6b9/lib/grape-swagger/endpoint.rb#L100

PR #811, opened 2020-10-09
Fixes #809: supports UTF8 route names.

### LINKED ISSUE #809: Can't generate doc for root when using international characters
When I use æøå in the root url, there is no generation of operations. 

Simple example:
```
require 'date'
require 'grape'
require 'grape-swagger'
require 'json'

module Grunnbeløp
  class G < Grape::API
    version 'v1', using: :path
    format :json
    prefix :api

    resource :grunnbeløp do
      desc 'returnerer grunnbeløp'
      get do
        {}
      end
    end

    add_swagger_documentation
  end
end
```

I'm guessing there is a regex looking for endpoints to document, and that one is only using `A-Za-z` or `/w`?

LeFnord:


LeFnord:
Closes #809

PR #815, opened 2021-01-06
Add required with multiple present

MaximeRDY:
Hello, 

I've added `required: true` for multiple present. 

Have a good day

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -27,7 +27,7 @@ group :development, :test do
   gem 'rake'
   gem 'rdoc'
   gem 'rspec', '~> 3.9'
-  gem 'rubocop', '1.6.1', require: false
+  gem 'rubocop', '~> 1.0', require: false
 end
 
 group :test do
--- a/Gemfile
+++ b/Gemfile
@@ -27,7 +27,7 @@ group :development, :test do
   gem 'rake'
   gem 'rdoc'
   gem 'rspec', '~> 3.9'
-  gem 'rubocop', '~> 1.0', require: false
+  gem 'rubocop', '1.6.1', require: false
 end
 
 group :test do
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 #### Features
 
 * Your contribution here.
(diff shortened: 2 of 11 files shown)


MaximeRDY:
@LeFnord Rubocop 1.7.0 is bugged so I've fixed to 1.6.1 (last version without bug)

LeFnord:
thanks @MaximeRDY  … somehow I can't figure it out, what require means in the context of a response … 🤔

Gemfile
```diff
@@ -27,7 +27,7 @@ group :development, :test do
   gem 'rake'
   gem 'rdoc'
   gem 'rspec', '~> 3.9'
-  gem 'rubocop', '~> 1.0', require: false
+  gem 'rubocop', '1.6.1', require: false
```
LeFnord, line 30: what is the bug?

MaximeRDY: They fix in Rubocop 1.8.0. So never mind :-)

MaximeRDY:
> thanks @MaximeRDY … somehow I can't figure it out, what require means in the context of a response … 🤔

In multiple present response, the response look like this: 
```
"responses": {
    "200": {
      "description": "Multiple response",
      "schema":{
        "type":"object",
        "properties":{
          "gender":{
            "$ref":"#/definitions/EnumValues"
          },
          "somethings":{
            "$ref":"#/definitions/Something"
          }
        }
      }
    }
  }
  ```
  
  But here the property `gender` and `somethings` is by definition optional (= can respond `null`) 
  So if you want to specify the property `somethings` cannot respond `null` we need to have a response like this : 
  ```
  "responses": {
    "200": {
      "description": "Multiple response",
      "schema":{
        "type":"object",
        "properties":{
          "gender":{
            "$ref":"#/definitions/EnumValues"
          },
          "somethings":{
            "$ref":"#/definitions/Something"
          }
        },
       "required": ["somethings"]
      }
    }
  }
  ```
  
  So it's why I've added the possibility to have something like :
  ```
  { model: Entities::Something, as: :somethings, is_array: true, required: true }
  ```

LeFnord:
ah ok @MaximeRDY … thanks for explanation … please add also a README entry and I'm happy to mörge it

LeFnord:
sorry, overseen it

PR #821, opened 2021-03-08
Parameter examples

### LINKED ISSUE #762: Example to param does not show up
I have been trying to add example to param but is does not show up in the swagger file.

```
params do
        requires :order_id, type: String,
                 documentation: {
                   type: 'String',
                   desc: 'Your unique order id of this order. (max 255 characters)',
                   example: '1234',
                   param_type: 'body'
                 }
end
```

But the swagger file only shows this:
```
...
"definitions":{ 
   "postOrders":{ 
      "type":"object",
      "properties":{ 
         "order_id":{ 
            "type":"string",
            "description":"Your unique order id of this order. (max 255 characters)"
         }
      }
   }
}
...
```

Any clues, workarounds or fixes?

dslh:
Hello. It would be useful for us to be able to specify example values for parameters, as per the [spec](https://swagger.io/docs/specification/2-0/adding-examples/). I'm not familiar with the internals of this project, let me know if I'm missing something in the implementation or tests. Thanks!

Closes #762.


dslh:
Hmmm, once I put this PR into action I find that the examples still aren't present in the resulting swagger file. Is there something I'm missing?

LeFnord:
thanks @dslh … to your question, I have no idea, your code and your spec are looking good, so maybe you want to investigate a bit more?

LeFnord:
@dslh … what did you think, should it be mörged and you investigate your issue in a separate PR … mayve it is not related to grape-swagger more to the used swagger client?

spec/swagger_v2/params_example_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+    Class.new(Grape::API) do
+      format :json
+
+      params do
+        requires :id, type: Integer, documentation: { example: 123 }
+        optional :name, type: String, documentation: { example: 'Person' }
+        optional :obj, type: 'Object', documentation: { example: { 'foo' => 'bar' } }
+      end
+
+      get '/endpoint_with_examples' do
```
LeFnord, line 16: @dslh … try with a post request ;)

PR #822, opened 2021-03-15
Corrected the related parameter lookup on request params.

Jack12816:
**- What is it good for**

There is a nasty bug while using grape-swagger on request parameters that share the same base name in any way (eg. `ids` and `user_ids`) which results in a missing parameter and a badly named on (eg. `ids` is missing, while `user_ids[][]` is renamed to). An example Grape API class looks like this:

```ruby
class Search < Grape::API
  params do
    with(allow_blank: false) do
      optional :ids, type: [String]
      optional :user_ids, type: [String]
    end
  end
  # [..]
end          
```

**- What I did**

I just changed the related parameter matching to use `start_with?` instead of `include?` to be sure that the related parameter is not accidentally matched while respecting actual matches. (eg. `user_id` and `user_id[id]`)

**- A picture of a cute animal (not mandatory but encouraged)**

download

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 #### Fixes
 
 * Your contribution here.
-
+* [#822](https://github.com/ruby-grape/grape-swagger/pull/822): Corrected the related parameter lookup on request params - [@Jack12816](https://github.com/Jack12816).
 
 ### 1.3.1 (November 1, 2020)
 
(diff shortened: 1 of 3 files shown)


LeFnord:
thank @Jack12816 … did it need an upgrade entry?

LeFnord:
ah … one more thing, please add an CHANGELOG.md entry

Jack12816:
> did it need an upgrade entry?
I don't think so, the expected behavior remains and this pr just fixes an edge case, but if you insist I'll can add an upgrade note :)

> one more thing, please add an CHANGELOG.md entry
done

LeFnord:
will update the branch, then doing the mörge

LeFnord:
thanks :)

Jack12816:
thank you, too! :) - what do you think about a release?

Jack12816:
@LeFnord ping :ping_pong:

PR #833, opened 2021-08-14
Examples in body params

stevenou:
When a parameter is `in: 'body'`, it appears to drop the example associated. So it will show: 
```
{
  email: 'string',
  password: 'string'
}
``` 
instead of 
```
{
  email: 'example@esp.com',
  password: 'wowthispasswordissosecure!'
}
```

I took a look at https://github.com/ruby-grape/grape-swagger/issues/762 and https://github.com/ruby-grape/grape-swagger/pull/821 which purportedly fix the issue, but it still wasn't working for me.

I'm really not familiar with this codebase, but I think the issue is within `GrapeSwagger::DocMethods::MoveParams` which has a `property_keys` that appears to limit which properties are retained when creating an "`in: body` object". I added the `example` key to that array and it seems to be working.

Would appreciate confirmation whether this is correct fix. Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 #### Fixes
 
-* Your contribution here.
+* [#833](https://github.com/ruby-grape/grape-swagger/pull/833): Fixes issue of examples not showing for `in: 'body'` parameters - [@stevenou](https://github.com/stevenou)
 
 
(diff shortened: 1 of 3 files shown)


LeFnord:
thanks @stevenou … and sorry for the late reply, please can you add a Changelog entry, thanks

stevenou:
@LeFnord done

PR #839, opened 2021-10-15
Fix documentation of `false`/`nil` default parameter values

magni-:
The `default` option currently gets dropped, since neither `false` nor `nil` are considered "present".


LeFnord:
ok thank you @magni-

PR #840, opened 2021-10-20
Fix documentation of `additionalProperties` field when used with array parameters

magni-:
This currently produces invalid Swagger:
```ruby
params do
  optional :array_param, type: Array, documentation: { additional_properties: true } do
    requires :nested_param
  end
end
```

I took the liberty of renaming `additionalProperties` to `additional_properties` (and deprecating the former) to have consistent casing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 #### Fixes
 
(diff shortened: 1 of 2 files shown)


magni-:
@LeFnord While working on adding documentation for this, I realized this will need more changes: OAS 2.0 [does _not_ support](https://github.com/OAI/OpenAPI-Specification/issues/668) `additionalProperties: true`. Is there a helper in the specs to ensure the generated JSON is valid OAS 2.0? I couldn't find one.

PR #844, opened 2021-11-18
Fixes the regexp used for parsing routes

senhalil:
When called `add_swagger_documentation` raises the following warning which is due to vertical bars `|` since they are parsed as literal characters and not as OR's in a character set `[]` definition.

```
grape-swagger-1.4.2/lib/grape-swagger.rb:35: warning: character class has duplicated range: /\/([\p{Alnum}|\-|\_]*?)[\.\/\(]/
```


senhalil:
It doesn't look like to me that the broken tests are due to my PR but I can fix them if that is the case. Let me know if that is the case please.

senhalil:
I just rebased the PR. Can someone give me some feedback? Afaiac, the regexp definition is not correct and needs a fix.

senhalil:
I just rebased the PR. Can someone give me some feedback -- maybe @LeFnord ?

LeFnord:
thanks @senhalil … makes sense to me

PR #847, opened 2022-02-24
Parse route_param type for nested endpoints

dmoss18:
We noticed that when we did this:
```rb
resource :accounts do
  route_param :account_number, type: String do
    resource :records do
      route_param :id do
        get do
          ...
        end
      end
    end
  end
end
```
The type of `account_number` is defaulted to `integer`. Grape doesn't include any `options` in `declared_params`, so this code parses any parent namespaces and builds a mapping `declared_params` to their `options`.


Gemfile
```diff
@@ -6,14 +6,14 @@ ruby RUBY_VERSION
 
 gemspec
 
-gem 'grape', case version = ENV['GRAPE_VERSION'] || '>= 1.5.0'
+gem 'grape', case version = ENV.fetch('GRAPE_VERSION', '>= 1.5.0')
```
LeFnord, line 9: please can you also adopt it to use the newest grape version: `'~> 1.6'`, thanks

LeFnord (commented):
Thanks @dmoss18 for the fix

please can you change the specs … maybe they are running only if grape '~> 1.6' is loaded or so

dmoss18:
@LeFnord I'm having a hard time getting Ruby version `3.2.0dev` installed locally. It seems to be downloading it straight from `ruby-dev-builds`. The best I could do is install 3.2.0-preview1 locally, and all spec tests passed with that version.

Do you have any tips I could try to get these tests to fail like they do in CI?

LeFnord:
thanks @dmoss18

PR #850, opened 2022-03-06
Fix value of `enum` to be `Array`

takahashim:
I am trying to use grape-swagger and other OpenAPI tools, but currently there seems to be a problem with the schema generated by grape-swagger.

In the OpenAPI 2.0 specification, `enum` uses the JSON Schema `enum`.

* https://github.com/OAI/OpenAPI-Specification/blob/f6f9ab3d88ad6ff0b235d5f4dc1a8ab28941d895/schemas/v2.0/schema.json#L1001-L1003
* https://swagger.io/specification/v2/

In JSON Schema, the value of an `enum` is expected to be an `Array`.

* https://datatracker.ietf.org/doc/html/draft-fge-json-schema-validation-00#section-5.5.1

However, in grape-swagger, an `enum` value can be a `String` literal.

This pull request fixes that and returns an `Array`.

I hope this change can be accepted.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 #### Fixes
 
+* [#850](https://github.com/ruby-grape/grape-swagger/pull/850): Fix value of `enum` to be `Array` - [@takahashim](https://github.com/takahashim)
 * Your contribution here.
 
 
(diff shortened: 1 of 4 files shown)


LeFnord:
thanks @takahashim … looks reasonable for me, please add an entry to the UPGRADE

takahashim:
Thanks a lot!  Also added a line to UPGRADING.md.

LeFnord (approved):

PR #852, opened 2022-03-10
Fix example to work without error

### LINKED ISSUE #851: `GrapeSwagger::Errors::UnregisteredParser` in example
When I run the code in example and try to get swagger_doc, I get an error.

## Steps to reproduce

```
$ ruby -v
ruby 2.7.5p203 (2021-11-24 revision f69aeb8314) [arm64-darwin21]
$ cd example
$ bundle install
$ bundle exec rackup
```

And when I accessed http://localhost:9292/, I got the following screen:

swagger_doc_error

It found similar issues #424, #548 and #693 . I understand you do not want to add `require "grape-swagger-entity"` explicitly.
If so, I think we can add Bundler.require. I am going to write a pull request later and would appreciate your review.

takahashim:
Fix #851.

* load `grape-swagger-entity` gem in `:development` environment
* require gem files with `Bundler.require`

I have a few concerns about this fix.

* We can remove other `require`s in `config.ru` such as `require 'rack/cors'`. Should I remove them?
* Instead of adding the `grape-swagger-entity` gem in `:development` to the common Gemfile, should I add another Gemfile for example as `example/Gemfile` ?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 * [#850](https://github.com/ruby-grape/grape-swagger/pull/850): Fix value of `enum` to be `Array` - [@takahashim](https://github.com/takahashim)
 * [#846] (https://github.com/ruby-grape/grape-swagger/pull/846): Fixes oapi rake tasks, allows generating sepcs for different API versions.
+* [#852](https://github.com/ruby-grape/grape-swagger/pull/852): Fix example to work without error - [@takahashim](https://github.com/takahashim)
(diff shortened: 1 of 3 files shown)


LeFnord:
Hi @takahashim … thanks for it, to your concerns:

1. it would be nice
2. think it is not needed, cause it is for development, the gem itself doesn't require it

takahashim:
Thank you! Fixed it and resolved the conflict.

LeFnord:
thanks @takahashim

PR #853, opened 2022-03-12
Add webrick to support Ruby 3.x

takahashim:
In Ruby 3.x, webrick is no longer included in the standard distribution.
So add it as a gem so that example will work properly.


LeFnord:
thanks again @takahashim … please can you update the changelog

takahashim:
Thank you!  I have rebased it.

LeFnord:
👍

PR #696, opened 2018-08-17
Required properties

Bugagazavr:
As mentioned https://github.com/ruby-grape/grape-swagger-entity/issues/34#issuecomment-412338781

Todo:

- [x] Grape swagger entity gem https://github.com/ruby-grape/grape-swagger-entity/pull/35
- [X] Grape swagger representable gem https://github.com/ruby-grape/grape-swagger-representable/pull/11
- [x] This PR

Release:

1. Release **grape-swagger** ( 0.31.0 )
2. Release a new **grape-swagger-entity** ( 0.3.0 ) and **grape-swagger-representable** ( 0.2.0 ), which depends on **grape-swagger** ( 0.31.0 )

PR #856, opened 2022-03-25
Remove unused methods in `GrapeSwagger::DocMethods::BuildModelDefinition`

takahashim:
I was reading the code for `GrapeSwagger::DocMethods::BuildModelDefinition` and noticed that `parse_entity()` and `parse_representable()` return only `nil`.
Then, since `required_attributes()` also returns only `nil`, the `if` statement in `build()` would also be unnecessary since `required_attrs.blank?` would always be true.

Based on this code reading, I've tried to remove the unused code.

### DIFF
--- a/lib/grape-swagger/doc_methods/build_model_definition.rb
+++ b/lib/grape-swagger/doc_methods/build_model_definition.rb
@@ -4,13 +4,8 @@ module GrapeSwagger
   module DocMethods
     class BuildModelDefinition
       class << self
-        def build(model, properties, required, other_def_properties = {})
+        def build(_model, properties, required, other_def_properties = {})
           definition = { type: 'object', properties: properties }.merge(other_def_properties)
-
-          if required.nil?
-            required_attrs = required_attributes(model)
-            definition[:required] = required_attrs unless required_attrs.blank?
-          end
 
           definition[:required] = required if required.is_a?(Array) && required.any?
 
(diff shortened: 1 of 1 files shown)


LeFnord:
mmh, a bit curious … do you checked it with [model-parsers](https://github.com/ruby-grape/grape-swagger#model-parsers-)?

takahashim:
Thanks for your comment. Yes, it is curious indeed.

It is a bit late, but I looked at the code and issues/PRs of grape-swagger-entity and grape-swagger-representable.
The trigger for this issue seemed to be a issue [Missing required attributes from merged entities](https://github.com/ruby-grape/grape-swagger-entity/issues/34) and a pull request [Required properties #696](https://github.com/ruby-grape/grape-swagger/pull/696) to resolve it.

In #696, `required` is added to the return value when `call`-ing `model_parsers`, and passed it to `GrapeSwagger::DocMethods::BuildModelDefinition.build()`.
Then, if `required` is nil, while falling back to the original process, a deprecated warning is output in `parse_entity` or `parse_representable`.

The deprecated warning added here is then removed in #768.

It is no problem that `deprecated_workflow_for` is removed in #768, but the `return unless` line above it should have been removed as well. The reason is that this line was intended to skip the subsequent warning and processing.
As a result, `parse_entity` and `parse_representable` are now methods that do nothing. And the `required_attributes` method that calls these two methods could have been removed as well.

LeFnord:
thanks @takahashim … good spot

PR #859, opened 2022-04-22
Fixes (restores) route ordering

### LINKED ISSUE #679: Grape swagger shows operations in reverse order.
After this [pull request](https://github.com/ruby-grape/grape-swagger/pull/631) has been merged,
grape swagger shows operations in reverse order from the order they were defined.
I think it makes browsing the documentation a bit more unnatural.

The test case that was provided in the above PR was a case where two apis were defined for same route, which I think is better served by using a versioning strategy.

dchandekstark:
This restores the route ordering which was changed in #631.

Fixes #679


dchandekstark:
Should I fix Rubocop issues unrelated to my changes?

dchandekstark:
I have one failing test, but I have to wonder if it's due to a bug in grape.  The purported solution in #631 may have simply masked the problem.  If in fact the issue is with grape-swagger I don't think the appropriate solution is to simply reverse the order of the route set.  In any event, the test case seems to me rather odd.  I couldn't find anything quite like it in grape's docs or specs (ofc I could well have missed it).

dchandekstark:
Please note that this change may reintroduce the issue that #631 attempted to solved, albeit IMO incorrectly.

spec/swagger_v2/mount_override_api_spec.rb
```diff
@@ -25,24 +25,13 @@ def app
...
     Class.new(Grape::API) do
-      mount new_api
       mount old_api
+      mount new_api
 
       add_swagger_documentation format: :json
     end
   end
 
-  context 'actual api request' do
```
LeFnord, line 35: fixing a spec by deleting it isn't so nice … 🤔

LeFnord: better you specify the the expected ordering so we have a documentation of …
it should be a bit easier then scrolling to many PRs … 🙂

dchandekstark: @LeFnord Thanks for taking a look. I share your concern about removing a test, but I'm rather unsure about how to reinstate it. The problem is that AFAICT Grape doesn't specify the behavior in this case -- i.e., two mounts at the same path with the same endpoint paths. I could just pick the one that makes the test pass, but that seems hardly better IMHO than removing the test.  Does that make sense?

LeFnord: mmh, what is with a spec with documents the behaviour you are introducing/correcting → the ordering of a route?

dchandekstark: @LeFnord The root problem is noted in #631 -- i.e., that grape-swagger is taking the last matching endpoint, whereas grape takes the first.  The code in question (https://github.com/ruby-grape/grape-swagger/blob/master/lib/grape-swagger.rb#L26-L45) is in a private method.  This problem is not correctly solved by simply reversing the order of the routes; something more sophisticated will be needed.  I don't have that solution in hand, but for my own usage I'm satisfied to simply not override mounts in order to avoid this issue.

dchandekstark:
I know that I have dodged a problem by removing a test (and I haven't yet tried adding it back), but this route ordering problem still exists.  I love this project, and I'd like to help fix this.  I have merged in the latest changes and all other tests are passing.

LeFnord:
thanks for finishing it
ok will merge it, will see if the ordering discussion pops up again ;)

LeFnord (approved):

PR #861, opened 2022-06-08
Fix rubocop warnings with auto correct

takahashim:
In the GitHub Actions for `grape-swagger`, there seems to be files that violates rubocop's cops, so I fixed them using the `--autocorrect` option.


LeFnord:
closed via #862

PR #862, opened 2022-07-14
Allow using nicknames for body definitions

magni-:
We have an API defined roughly as follows:

```ruby
resource :things do
  desc "Batch update things" do
    nickname "put-things"
  end
  params do
    # a bunch of params
  end
  put "/" do
    # ...
  end

  segment "/:id" do
    desc "Update thing" do
      nickname "put-thing"
    end
    params do
      # a bunch of other params
    end
    put "/" do
      # ...
    end
  end
end
```

We noticed the docs for our API were incorrect. Both `paths.things.put.parameters.0.schema.$ref` and `paths.things{id}.put.parameters.1.schema.$ref` (`parameters.0` is for the path `:id` parameter) are pointing to the same thing: `"#/definitions/putThings`. It turns out the `put-thing` body parameter definition is overriding the previous `put-things` body parameter definition, rather than defining its own thing. It's been a while since I've dug into the `grape-swagger` code, so I went with the (IMO) easy fix: using route nicknames if they're available to name the body parameter definitions.

Edit: Based on the changes I had to make to get specs to pass, this change would also have solved our issue without using route nicknames, since the references become `putTests` and `putTestsId`.

Related issue: https://github.com/ruby-grape/grape/issues/2035

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,10 +3,10 @@
 #### Fixes
 
 * [#850](https://github.com/ruby-grape/grape-swagger/pull/850): Fix value of `enum` to be `Array` - [@takahashim](https://github.com/takahashim)
-* [#846] (https://github.com/ruby-grape/grape-swagger/pull/846): Fixes oapi rake tasks, allows generating sepcs for different API versions.
+* [#846](https://github.com/ruby-grape/grape-swagger/pull/846): Fixes oapi rake tasks, allows generating sepcs for different API versions.
 * [#852](https://github.com/ruby-grape/grape-swagger/pull/852): Fix example to work without error - [@takahashim](https://github.com/takahashim)
-* Your contribution here.
 * [#853](https://github.com/ruby-grape/grape-swagger/pull/853): Add webrick gem so that example works in Ruby 3.x - [@takahashim](https://github.com/takahashim)
+* Your contribution here.
 
 ### 1.4.2 (October 22, 2021)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 * [#846](https://github.com/ruby-grape/grape-swagger/pull/846): Fixes oapi rake tasks, allows generating sepcs for different API versions.
 * [#852](https://github.com/ruby-grape/grape-swagger/pull/852): Fix example to work without error - [@takahashim](https://github.com/takahashim)
 * [#853](https://github.com/ruby-grape/grape-swagger/pull/853): Add webrick gem so that example works in Ruby 3.x - [@takahashim](https://github.com/takahashim)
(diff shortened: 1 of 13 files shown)


magni-:
I'm still seeing the same failure on `mustermann@2.0.1`. It looks like it would work if we added `nil` before `constraint` here: https://github.com/ruby-grape/mustermann-grape/blob/v1.0.2/lib/mustermann/grape.rb#L20 ? But the proper fix seems to be an update to `mustermann` itself, I started a PR here: https://github.com/sinatra/mustermann/pull/134

Either way I don't think this should be a blocker for this PR 🙇🏼

magni-:
@LeFnord anything else I should add to this PR, or does this look good to go? 🙇🏼 Thanks!

magni-:
> I'm still seeing the same failure on `mustermann@2.0.1`. It looks like it would work if we added `nil` before `constraint` here: https://github.com/ruby-grape/mustermann-grape/blob/v1.0.2/lib/mustermann/grape.rb#L20 ? But the proper fix seems to be an update to `mustermann` itself, I started a PR here: [sinatra/mustermann#134](https://github.com/sinatra/mustermann/pull/134)
> 
> Either way I don't think this should be a blocker for this PR 🙇🏼

`mustermann v2.0.2` [was released](https://github.com/sinatra/mustermann/releases/tag/v2.0.2) and the build is now fully green 🎉

LeFnord:
thanks @magni- … it seems it changes something so maybe add an entry to UPGRADING.md

and one question: should we add `gem musterman, '~> 2.0.2'` to ensure it works correct?

magni-:
> it seems it changes something so maybe add an entry to UPGRADING.md

👍🏼 

> and one question: should we add `gem musterman, '~> 2.0.2'` to ensure it works correct?

Good question. The build works fine on all released versions of Ruby, it's just the upcoming 3.2 (which won't be released for another 5 months) which had issues. If we add an explicit requirement, might as well make it to `~> 3` which was just released (the only change from `2.0.2` is dropping support for Ruby 2.5 and below. `grape-swagger` already requires Ruby 2.7+).

magni-:
@LeFnord I've updated `UPGRADING`. For the version number, this feels more like a `1.5.0` than a `1.4.3` ?

LeFnord:
@magni- … yeah, definitely will it be 1.5.0 …
what is with `mustermann`, should it be aded to the Gemfile?

magni-:
> @magni- … yeah, definitely will it be 1.5.0 … what is with `mustermann`, should it be aded to the Gemfile?

I think it makes more sense to update the requirement in `mustermann-grape` instead? https://github.com/ruby-grape/mustermann-grape/blob/master/mustermann-grape.gemspec#L20

LeFnord:
ok makes sense :)

magni-:
@LeFnord could you release this to RubyGems? Thank you! 🙇🏼

LeFnord:
done

Issue #677, opened 2018-05-27
Think about add the desc options 'consumes' or 'produces' to add_swagger_documentation?


LuckyScott:
Can you add the API desc options 'consumes' or 'produces' to add_swagger_documentation?
So that I don't need to add option 'consumes' to every API desc.

spaceraccoon:
Hey folks, I've created a PR that adds a global option at #872. Hope it gets reviewed/added soon.

Issue #623, opened 2017-07-24
Default consumes setting for desc method


jeyraof:
My POST apis almost do with formData.
Now, these methods have consumes setting manually:

```ruby
desc 'method1', consumes: ['application/x-www-form-urlencoded']
params {}
post {}

desc 'method3', consumes: ['application/x-www-form-urlencoded']
params {}
post {}

...
```

Is the way to set default consumes to ` ['application/x-www-form-urlencoded']` ?

pdkproitf:
I have to do the same thing when I try to integrate Swagger-ui into the project.

jackalcooper:
Failed to find a way to make it global. Here is my walk around for anyone who are interested.
Add this field to your `SwaggerUIBundle` parameter
```javascript
        requestInterceptor: function (request) {
          request.headers["Content-Type"] = "application/x-www-form-urlencoded; charset=UTF-8"
          return request
        }
```

spaceraccoon:
Hey folks, I've created a PR that adds a global option at #872. Hope it gets reviewed/added soon.

PR #872, opened 2022-10-07
Add consumes and produces options to add_swagger_documentation

spaceraccoon:
As per issues #677 and #623, there is demand to allow for setting the global/default `consumes` and `produces` field value via `add_swagger_documentation`. In any case, the global fields are officially supported in the [OpenAPI spec](https://swagger.io/specification/v2/#swaggerConsumes).

This will make it a lot easier to support non-JSON content-types at the global level, since the current default is `application/json`, forcing affected users to set the type at each endpoint instead of globally/by default.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 #### Features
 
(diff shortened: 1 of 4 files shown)


spaceraccoon:
@LeFnord wondering if you are interested in reviewing this since you mentioned this is a welcome feature in #677 :) Great contributing guide btw!

LeFnord:
@spaceraccoon … sorry for late reply … yes this feature is welcome :) … thank you

PR #877, opened 2022-11-08
Search for routes that contain swagger_doc

igor-drozdov:
If a client-defined route contains `doc` word, then the call to that URL will be unexpectedly performed

Let's filter the routes using `swagger_doc` instead

### DIFF
--- a/lib/grape-swagger/rake/oapi_tasks.rb
+++ b/lib/grape-swagger/rake/oapi_tasks.rb
@@ -95,7 +95,7 @@ module GrapeSwagger
       def urls_for(api_class)
         api_class.routes
                  .map(&:path)
-                 .select { |e| e.include?('doc') }
+                 .select { |e| e.include?('swagger_doc') }
                  .reject { |e| e.include?(':name') }
                  .map { |e| format_path(e) }
                  .map { |e| [e, ENV.fetch('resource', nil)].join('/').chomp('/') }


anothermh:
@igor-drozdov Thank you for working on this. I am having a problem with this method at the moment as well and it's an edge case you may want to consider in your fix here.

The default value for `mount_path` is `/swagger_doc` but this can be overridden in the call to `add_swagger_documentation`:

https://github.com/ruby-grape/grape-swagger/blob/9fda24bb51ec8fc2d4f58ebc6e7e68692d7d042a/lib/grape-swagger.rb#L44

I override this with a custom path and as a result I can't use the rake task to export documentation because `/swagger_doc` doesn't exist in my routes.

I'm overriding this method for now:

```ruby
require 'rake/tasklib'

module GrapeSwagger
  module Rake
    class OapiTasks < ::Rake::TaskLib
      private

      def urls_for(api_class)
        api_class.routes
          .map(&:path)
          .select { |e| e.include?('/open_api') }
          .reject { |e| e.include?(':name') }
          .map { |e| format_path(e) }
          .map { |e| [e, ENV.fetch('resource', nil)].join('/').chomp('/') }
      end
    end
  end
end
```

If you're able to capture the `mount_path` and use that instead of `swagger_doc` that would resolve the bug I'm experiencing. I don't know if the value can be read, this has come up only in the last few hours.

igor-drozdov:
> If you're able to capture the mount_path and use that instead of swagger_doc that would resolve the bug I'm experiencing. I don't know if the value can be read, this has come up only in the last few hours.

@anothermh thanks for your feedback!

Since the variable is configurable, then my _fix_ is a breaking change because someone can have `openapi_doc` and it will stop working for them 🤔 We definitely need to capture the `mount_path` value but I'm also not sure that it can be done. We can path the mount path as a param to the rake task though but have this value `doc` by default for backward compatibility.

Let's see what @LeFnord thinks about it 👍

LeFnord:
sorry @igor-drozdov …
> Since the variable is configurable, then my _fix_ is a breaking change because someone can have `openapi_doc` and it will stop working for them 🤔 We definitely need to capture the `mount_path` value but I'm also not sure that it can be done. We can path the mount path as a param to the rake task though but have this value `doc` by default for backward compatibility.

yes that is exactly the case, mount_path is configurable, so your change can/will break 
so try instead to fetch the mount path from the grape-swagger configuration

olivier-thatch:
@numbata This PR chan be closed, #924 fixed this issue.

Issue #721, opened 2018-10-30
Swagger-ui for JSON APIs sends incorrect content-type 


dblock:
Coming from https://github.com/swagger-api/swagger-ui/issues/4981

Swagger UI produces multipart bodies on POST tries. I believe this is because the swagger API produces fields with `in: 'formData'`. It seems that the code does that for any primitive type, instead of doing `in: 'body'`. Doesn't seem right.

Monkey patching this works around the problem.

```ruby
module GrapeSwagger
  module DocMethods
    class DataType
      class << self
        def request_primitive?(type)
          false
        end
      end
    end
  end
end
```

But probably not the right solution?

hardywu:
A related issue #623.  It should be either setting default consume to 'application/x-www-form-urlencoded', or setting  default params_type to 'body'.  Otherwise, Swagger-UI will be unusable without a patch.

mcfoton:
Have the same issue, would love to see this resolved 👀

henrytseng:
Just a quick ping on this issue.  Looks like it should be following: https://swagger.io/docs/specification/describing-request-body/

tristil:
My workaround FYI is to add `documentation: { param_type: 'body' }` like this:

```ruby
   requires(
            :data,
            type: Array[JSON],
            documentation: { param_type: 'body' }
          ) do
  end
```

tristil:
Relevant lines are 

https://github.com/ruby-grape/grape-swagger/blob/564b486430881a68fef0cb12dcff583503fd8911/lib/grape-swagger/doc_methods/parse_params.rb#L67

and 

https://github.com/ruby-grape/grape-swagger/blob/564b486430881a68fef0cb12dcff583503fd8911/lib/grape-swagger/doc_methods/parse_params.rb#L89

spaceraccoon:
Hi all, I've made a PR fixing this default behaviour at #880

dhruvCW:
fixed in #927

PR #880, opened 2022-11-11
Set default parameter location based on consumes

spaceraccoon:
As per #721, `grape-swagger` sets the default `consumes` to `['application/json']` in `lib/grape-swagger/doc_methods/produces_consumes.rb`:

```
return ['application/json'] unless args.flatten.present?
```

However, it also sets the default `in` of parameters to `formData` in `lib/grape-swagger/doc_methods/parse_params.rb`:

```
...
  @parsed_param[:in] = param_type || 'formData'

def param_type(value_type)
...
  elsif %w[POST PUT PATCH].include?(value_type[:method])
    DataType.request_primitive?(value_type[:data_type]) ? 'formData' : 'body'
```

Here, `DataType.request_primitive` simply checks if the parameter's data type is one of `integer long float double byte date dateTime binary password email uuid object string boolean file json array`, in which case its location is set to `formData`. As such, this leads to a situation where most parameters are set to `formData` by default while their endpoint is also set to `consumes` `application/json` by default:

```
      consumes:
        - application/json
      parameters:
        - in: formData
```

However, as described in the [OpenAPI V2 documentation](https://swagger.io/specification/v2/#parameter-object) (also [here](https://swagger.io/docs/specification/2-0/describing-request-body/)), `the payload of the application/x-www-form-urlencoded and multipart/form-data requests is described by using form parameters, not body parameters` and `formData` should only be used for `application/x-www-form-urlencoded` and `multipart/form-data` requests. Currently, the default output of `grape-swagger` will lead to OpenAPI validation errors.

This MR adds logic to set the default parameter location (`in`) based on the `consumes` object, setting it to `formData` if `consumes` is `application/x-www-form-urlencoded` or `multipart/form-data` and `body` if otherwise for `PUT POST PATCH` requests. This can still be overriden by endpoint-specific options as is the usual behaviour.

However, since several specs relied on this invalid output previously, it is necessary to make some changes to ensure passing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 #### Features
 
+* [#880](https://github.com/ruby-grape/grape-swagger/pull/880): Set default parameter location based on consumes - [@spaceraccoon](https://github.com/spaceraccoon)
 * Your contribution here.
 
 #### Fixes
--- a/lib/grape-swagger/doc_methods/parse_params.rb
+++ b/lib/grape-swagger/doc_methods/parse_params.rb
@@ -4,7 +4,7 @@ module GrapeSwagger
   module DocMethods
     class ParseParams
       class << self
-        def call(param, settings, path, route, definitions)
+        def call(param, settings, path, route, definitions, consumes)
           method = route.request_method
           additional_documentation = settings.fetch(:documentation, {})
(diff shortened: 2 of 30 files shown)


dhruvCW:
this can be closed since the changes have been merged with #927

PR #895, opened 2023-03-25
Fix README links

duffn:
- Fix redirecting GitHub links with current org names
- Fix old online demo link


LeFnord:
thanks @duffn

PR #903, opened 2023-07-06
Assign :example param on array properties

VladMomotov:


LeFnord (commented):
@VladMomotov … 👍and sorry for late reply

please can you add an CHANGELOG entry and  
rebase against master, it fixes the rubocop stuff

LeFnord (approved):
Thanks @VladMomotov

Issue #904, opened 2023-09-05
Should the minimum, maximum values be only Integer type?


artyomb:
https://github.com/ruby-grape/grape-swagger/blob/3a5da56fb3ad19dcb07cf3ef86549bb811e22dcc/lib/grape-swagger/doc_methods/parse_params.rb#L165C67-L165C67

As described in https://docs.swagger.io/spec.html

> **minimum** | string | **number**, integer | The minimum valid value for the type, inclusive. If this field is used in conjunction with the defaultValue field, then the default value MUST be higher than or equal to this value. The value type is string and should represent the minimum numeric value. Note: This will change to a numeric value in the future.

The values may be number (float or double). Should be the validation be removed?
```ruby
 ... if values.first.is_a?(Integer) 
```

PR #905, opened 2023-09-05
Allow range be Float

artyomb:
https://github.com/ruby-grape/grape-swagger/issues/904#issue-1881746890

As described in https://docs.swagger.io/spec.html

minimum | string | number, integer | The minimum valid value for the type, inclusive. If this field is used in conjunction with the defaultValue field, then the default value MUST be higher than or equal to this value. The value type is string and should represent the minimum numeric value. Note: This will change to a numeric value in the future.

The values may be number (float or double)?

### DIFF
--- a/lib/grape-swagger/doc_methods/parse_params.rb
+++ b/lib/grape-swagger/doc_methods/parse_params.rb
@@ -162,7 +162,7 @@ module GrapeSwagger
           when Proc
             parse_enum_or_range_values(values.call) if values.parameters.empty?
           when Range
-            parse_range_values(values) if values.first.is_a?(Integer)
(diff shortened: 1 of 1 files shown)


LeFnord:
@artyomb … yeah, makes sense to allow any kind of number

please rebase against master, it fixes the rubocop stuff

artyomb:
Sorry, I broke something. I don't know how to rebase with this pull request.

artyomb:
I did "sync fork". Hope it helped?

LeFnord:
no – please have a look on the failures, it seems your change is related to it

artyomb:
Ok, sorry. I don't know how complete the request. Please close/reject it. Maybe move to the issues.

artyomb:
The patch is still actual, is it possible to accept the request?

PR #907, opened 2023-10-22
Refactor gemspec files to include just the necessary

ericproulx:
Hi, while installing the gems, I've noticed that spec files are included when installed. I've refactored the `files` from your gemspec just to include the necessary.

### DIFF
--- a/grape-swagger.gemspec
+++ b/grape-swagger.gemspec
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
-$LOAD_PATH.push File.expand_path('lib', __dir__)
-require 'grape-swagger/version'
+require_relative 'lib/grape-swagger/version'
 
 Gem::Specification.new do |s|
   s.name        = 'grape-swagger'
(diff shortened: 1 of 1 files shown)


grape-swagger.gemspec
```diff
@@ -18,6 +17,6 @@ Gem::Specification.new do |s|
   s.required_ruby_version = '>= 2.7'
   s.add_runtime_dependency 'grape', '~> 1.3'
 
-  s.files         = `git ls-files`.split("\n")
+  s.files = Dir['lib/**/*', 'CHANGELOG.md', 'CONTRIBUTING.md', 'README.md', 'RELEASING.md', 'UPGRADING.md', 'LICENSE.txt', 'grape-swagger.gemspec']
```
LeFnord, line 20: wouldn't it be more explicit by excluding the spec dir?

ericproulx: Screenshot 2023-10-22 at 14 29 43
I think it's easier to include specific files than rejecting. Yet, bundler [template](https://github.com/rubygems/rubygems/blob/812366203b3fd9bb4e278db790c23f35f947c733/bundler/lib/bundler/templates/newgem/newgem.gemspec.tt#L30-L33) is rejecting folders and files but not all of them. For instance, the Dangerfile would still be there, even the .github folder.

ericproulx: No that's just the template applied if you `bundler gem` today

LeFnord (approved):

PR #909, opened 2023-10-24
grape-swagger needs rack-test

### LINKED ISSUE #908: Should rack-test be considered an optional dependency?
I'm running `swagger-ui` in a docker container by [providing it a `swagger.json` file][a]. Which I build with [`rake oapi:fetch`][d]. After switching from `bundle install` to `bundle install --without test development` I realized that `grape-swagger` [needs `rack-test`][c], which is [not in the dependencies][b]. Do you want to add it to the dependencies? Or should I install it manually? Or I'm doing something wrong? On a side note, why does the `rake` task need `rack-test`?

[a]: https://github.com/swagger-api/swagger-ui/blob/v5.9.0/docs/usage/installation.md#docker
[b]: https://rubygems.org/gems/grape-swagger/versions/1.6.1
[c]: https://github.com/ruby-grape/grape-swagger/blob/v1.6.1/lib/grape-swagger/rake/oapi_tasks.rb#L5
[d]: https://github.com/ruby-grape/grape-swagger/blob/v1.6.1/README.md#rake

x-yuri:
Fixes #908

### DIFF
--- a/grape-swagger.gemspec
+++ b/grape-swagger.gemspec
@@ -16,6 +16,7 @@ Gem::Specification.new do |s|
 
   s.required_ruby_version = '>= 2.7'
   s.add_runtime_dependency 'grape', '~> 1.3'
+  s.add_runtime_dependency 'rack-test', '~> 2'
 
   s.files = Dir['lib/**/*', '*.md', 'LICENSE.txt', 'grape-swagger.gemspec']
   s.require_paths = ['lib']


x-yuri:
I specified `~> 2` because it's kind of optimistic not to provide a version constraint, but maybe it can work with older versions?..

LeFnord:
thanks @x-yuri … think it should be fine with version `~>2`

PR #910, opened 2023-11-13
Allow using Grape v2

ninoseki:
Allow using Grape v2.

As far as I know, v2 does not introduce any breaking changes to this gem. So it's safe to relax the version requirement.

### DIFF
--- a/grape-swagger.gemspec
+++ b/grape-swagger.gemspec
@@ -15,7 +15,7 @@ Gem::Specification.new do |s|
   s.metadata['rubygems_mfa_required'] = 'true'
 
   s.required_ruby_version = '>= 2.7'
-  s.add_runtime_dependency 'grape', '~> 1.3'
+  s.add_runtime_dependency 'grape', '>= 1.3', '< 3.0'
   s.add_runtime_dependency 'rack-test', '~> 2'
 
   s.files = Dir['lib/**/*', '*.md', 'LICENSE.txt', 'grape-swagger.gemspec']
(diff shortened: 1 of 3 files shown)


grape-swagger.gemspec
```diff
@@ -15,7 +15,7 @@ Gem::Specification.new do |s|
   s.metadata['rubygems_mfa_required'] = 'true'
 
   s.required_ruby_version = '>= 2.7'
-  s.add_runtime_dependency 'grape', '~> 1.3'
+  s.add_runtime_dependency 'grape', '>= 1.3', '< 3.0'
```
LeFnord, line 18: why?

LeFnord (commented):
thanks @ninoseki … please add an Changelog entry

grape-swagger.gemspec
mscrivo: because grape is now version 2.0.0. `~> 1.3` explicitly means >= 1.3, < 2.0

mscrivo: If you reference both these gems, you cannot update to grape 2.0 because of this constraint

ninoseki:
Thanks for checking. I updated CHANGELOG.md. Please let me know if I need to update something other.

mscrivo:
Any plans to publish a release soon @LeFnord? It's cool if you're not, just thought I'd check before pointing to a github commit to get this change.

LeFnord:
wanted to release it today eveneing CET

LeFnord:
done, released 2.0.0

Issue #915, opened 2024-01-02
Restore Ruby 3.0 support?


godfat:
GitLab attempted to upgrade `grape-swagger` from `1.6.1` to `2.0.0`, however GitLab has to revert it because 2.0.0 doesn't support Ruby 3.0: https://gitlab.com/gitlab-org/quality/engineering-productivity/master-broken-incidents/-/issues/4841#note_1710824588

I noticed this was dropped at https://github.com/ruby-grape/grape-swagger/commit/8e723c6383e40ea47b6a6aad3391e98bbd356cad

And I see the reasoning was:

> * Aligns ruby and grape version with test matrix.

Which makes sense to me. However, I would like to ask if it's possible to at least restore support for Ruby 3.0 if it's actually working. I tried to run tests locally with Ruby 3.0 and everything passed.

godfat:
I created a pull request for consideration: https://github.com/ruby-grape/grape-swagger/pull/916

LeFnord:
closed via #916 

and released as 2.0.1

PR #916, opened 2024-01-02
Restore Ruby 3.0 support

godfat:
All tests passed locally. See https://github.com/ruby-grape/grape-swagger/issues/915

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,7 +30,7 @@ jobs:
       GRAPE_VERSION: '1.7.1'
     strategy:
       matrix:
-        ruby-version: ['3.1', '3.2', '3.3', 'head']
+        ruby-version: ['3.0', '3.1', '3.2', '3.3', 'head']
     steps:
     - name: Check out branch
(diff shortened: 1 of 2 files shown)


LeFnord:
does it make sense to also readd it to CI matrix, or are you fine with that?

LeFnord:
@godfat ^^^

godfat:
@LeFnord Good question. Thanks for asking. It'll be the best to add it to CI matrix as well. However, I see adding it can increase a lot of cost to CI because there's some combinations of versions. If this is acceptable, that would be great. If it's not, then I think it's fine to support it without actually running CI, especially given that it's not being run for some time and it's still passing?

What do you think?

LeFnord:
mmh … think I'll mörge it, and adapt the CI matrix afterwards, also make a release

godfat:
I amended the commit and added 3.0 support to CI as well. It's very fast. Maybe the cost is fine :)

godfat:
Awesome, thank you so much! A new release will be great, then we can upgrade very quickly.

Issue #666, opened 2018-02-14
Array of entities in request body


yewton:
Related issue: https://github.com/ruby-grape/grape-entity/issues/252

As for example above, we want request body to be:

```json
{
  "accounts": [
    {
      "cma": 0,
      "name": "string",
      "environment": "string",
      "sites": 0,
      "username": "string",
      "password": "string"
    }
  ]
}
```

But actually:

```json
[
    {
      "cma": 0,
      "name": "string",
      "environment": "string",
      "sites": 0,
      "username": "string",
      "password": "string"
    }
]
```

Grape also wants the array to be in the key of top level JSON object.
( Top level array data cannot be handled by Grape out of the box ( cf. https://github.com/ruby-grape/grape/issues/1730 ). )

So it seems the generated specification is broken.

A workaround is to adding dummy optional parameter.
It forces the request body to be a JSON object.

I've found that making [`MoveParams#build_definition`](https://github.com/ruby-grape/grape-swagger/blob/v0.28.0/lib/grape-swagger/doc_methods/move_params.rb#L164) to always use `object_type` fixes the issue.
Is this reasonable fix?

lstanden:
I just ran into the same bug.  Really don't want to go the dummy parameter route, but looks like it's not possible right now.

urkle:
The issue seems to be with the `correct_array_param`/`should_correct_array?` method in move_params.b

I have this param contract.

```ruby
requires :members, type: Array, documentation: {param_type: 'body'} do
  requires :user_id, type: Integer, desc: 'User ID'
  requires :role, type: Symbol, values: [:viewer, :user, :manager, :remove], desc: 'Role for the user.'
end
```

And it yields incorrect swagger docs expecting this.
```json
[
  {
    "members": [
      {
        "user_id": 0,
        "role": "string"
      }
    ]
  }
]    
```

When it should be like this.

```json
  {
    "members": [
      {
        "user_id": 0,
        "role": "string"
      }
    ]
  }
```

swistaczek:
Hey, did you guys find out a workaround for the problem mentioned in this issue?

urkle:
@swistaczek what I have been doing to work around this issue is the following.
```ruby
params do
  requires :members, type: Array, documentation: {param_type: 'body'} do
    requires :user_id, type: Integer, desc: 'User ID'
    requires :role, type: Symbol, values: [:viewer, :user, :manager, :remove], desc: 'Role for the user.'
   end
[...]

PR #922, opened 2024-04-11
Force request body to be an schema object

numbata:
## Motivation

There is an ongoing issue in grape-swagger where array parameters are incorrectly generated as top-level arrays in the Swagger documentation, as described in GitHub issue [#666](https://github.com/ruby-grape/grape-swagger/issues/666). This misrepresentation leads to API integration issues and confusion, as grape expects arrays to be part of a JSON object in request bodies. The community has resorted to workarounds such as introducing dummy parameters to force the correct structure, which is not ideal for clean and maintainable code.

## Solution

This PR addresses the issue by enforcing that all request bodies are defined as schema objects within the Swagger documentation, even if they contain arrays. By modifying how grape-swagger interprets and documents array types, this PR ensures that arrays are always nested within an object schema, aligning Swagger output with grape requirements and common API design practices. The changes eliminate the need for workarounds, sand prevent potential integration issues, making the API documentation clearer and more accurate.

### DIFF
--- a/lib/grape-swagger/doc_methods/move_params.rb
+++ b/lib/grape-swagger/doc_methods/move_params.rb
@@ -18,8 +18,6 @@ module GrapeSwagger
 
           params_to_move = movable_params(params)
 
-          return (params + correct_array_param(params_to_move)) if should_correct_array?(params_to_move)
-
           params << parent_definition_of_params(params_to_move, path, route)
 
           params
@@ -27,21 +25,11 @@ module GrapeSwagger
 
         private
 
-        def should_correct_array?(param)
-          param.length == 1 && param.first[:in] == 'body' && param.first[:type] == 'array'
-        end
-
-        def correct_array_param(param)
-          param.first[:schema] = { type: param.first.delete(:type), items: param.first.delete(:items) }
-
-          param
-        end
-
         def parent_definition_of_params(params, path, route)
           definition_name = OperationId.build(route, path)
-          build_definition(definition_name, params)
+          # NOTE: Parent definition is always object
+          @definitions[definition_name] = object_type
           definition = @definitions[definition_name]
-
           move_params_to_new(definition, params)
 
           definition[:description] = route.description if route.try(:description)
@@ -53,6 +41,7 @@ module GrapeSwagger
           params, nested_params = params.partition { |x| !x[:name].to_s.include?('[') }
           params.each do |param|
             property = param[:name]
+
             param_properties, param_required = build_properties([param])
             add_properties_to_definition(definition, param_properties, param_required)
             related_nested_params, nested_params = nested_params.partition { |x| x[:name].start_with?("#{property}[") }
--- a/lib/grape-swagger/doc_methods/parse_params.rb
+++ b/lib/grape-swagger/doc_methods/parse_params.rb
@@ -70,7 +70,7 @@ module GrapeSwagger
 
         def document_array_param(value_type, definitions)
           if value_type[:documentation].present?
-            param_type = value_type[:documentation][:param_type]
+            param_type = value_type[:documentation][:param_type] || value_type[:documentation][:in]
             doc_type = value_type[:documentation][:type]
             type = DataType.mapping(doc_type) if doc_type && !DataType.request_primitive?(doc_type)
             collection_format = value_type[:documentation][:collectionFormat]
--- a/spec/issues/539_array_post_body_spec.rb
+++ b/spec/issues/539_array_post_body_spec.rb
@@ -39,19 +39,35 @@ describe '#539 post params given as array' do
   let(:definitions) { subject['definitions'] }
 
   specify do
-    expect(parameters).to eql(
+    expect(parameters).to match(
       [
         {
-          'in' => 'body', 'name' => 'elements', 'required' => true, 'schema' => {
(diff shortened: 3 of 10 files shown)


lib/grape-swagger/doc_methods/parse_params.rb
```diff
@@ -70,7 +70,7 @@ def document_add_extensions(settings)
 
         def document_array_param(value_type, definitions)
           if value_type[:documentation].present?
-            param_type = value_type[:documentation][:param_type]
+            param_type = value_type[:documentation][:param_type] || value_type[:documentation][:in]
```
numbata, line 73: This seems to be the root cause of problems in recognizing a "body"/"formData" type. Currently, it's possible to mark a parameter as a body parameter with any of these options.

spec/issues/539_array_post_body_spec.rb
```diff
@@ -39,19 +39,35 @@ class ArrayOfElements < Grape::Entity
   let(:definitions) { subject['definitions'] }
 
   specify do
-    expect(parameters).to eql(
+    expect(parameters).to match(
       [
         {
-          'in' => 'body', 'name' => 'elements', 'required' => true, 'schema' => {
```
numbata, line 45: This moved as-is to the definition of the request body schema.

spec/issues/542_array_of_type_in_post_body_spec.rb
```diff
@@ -25,22 +25,37 @@
...
 
   let(:parameters) { subject['paths']['/issue_542']['post']['parameters'] }
+  let(:definitions) { subject['definitions'] }
 
   specify do
     expect(parameters).to eql(
       [
         {
           'in' => 'body',
-          'name' => 'logs',
```
numbata, line 34: This moved as-is to the definition of the request body schema.

spec/issues/553_align_array_put_post_params_spec.rb
```diff
@@ -96,26 +96,38 @@
...
       describe 'post request' do
         let(:params) { subject['paths']['/in_body']['post']['parameters'] }
+        let(:definitions) { subject['definitions'] }
 
         specify do
           expect(params).to eql(
             [
               {
                 'in' => 'body',
-                'name' => 'guid',
```
numbata, line 105: In request schema definition now.

spec/swagger_v2/params_array_spec.rb
```diff
@@ -118,8 +118,7 @@
         end
 
         specify do
-          expect(subject['definitions']['postArrayOfType']['type']).to eql 'array'
-          expect(subject['definitions']['postArrayOfType']['items']).to eql(
```
numbata, line 122: This confused me:
```
          # as body parameters it would be interpreted a bit different,
          # cause it could not be distinguished anymore, so this would be translated to one array,
          # see also next example for the difference
          params do
            requires :array_of_string, type: Array[String], documentation: { param_type: 'body', desc: 'nested array of strings', example: %w[a b] }
            requires :array_of_integer, type: Array[Integer], documentation: { param_type: 'body', desc: 'nested array of integers' }
          end
```
I still do not understand why this had an array type in a root and it was "okay". 🤔

lib/grape-swagger/doc_methods/move_params.rb
```diff
@@ -18,30 +18,18 @@ def to_definition(path, params, route, definitions)
...
-          param.first[:schema] = { type: param.first.delete(:type), items: param.first.delete(:items) }
-
-          param
-        end
-
         def parent_definition_of_params(params, path, route)
           definition_name = OperationId.build(route, path)
-          build_definition(definition_name, params)
+          # NOTE: Parent definition is always object
+          @definitions[definition_name] = object_type
```
numbata, line 31: Since it is [impossible](https://github.com/ruby-grape/grape/issues/1730) (in terms of effort/benefit) for the grape to have an array as a root object, I think this is an "okay"-ish change.

numbata:
@LeFnord Can I ask you to take a look at the PR?

LeFnord (commented):
yeah … I checked it out an tested it, works perfectly well, many thanks

please can you add a CHANGELOG entry, then I'll mörge and release it

LeFnord:
… hava a look on it → https://github.com/ruby-grape/grape-swagger/blob/master/CHANGELOG.md 😉

spec/swagger_v2/params_array_spec.rb
LeFnord: if I remember me correctly, then grape allows to define an Array of type in first place, but as you described, this is not the way OAPI handle it

means grape allows much more then OAPI2.x and we have to deal with it, it will change with support of OAPI3.x, but there is no progress atm

LeFnord: sorry, we wrote at the same time …

so think adding also a comment into the UPGRADE.md to let others know, that there is a possible breaking change

numbata: OAPI3.x 😍

numbata:
@LeFnord CHANGELOG and UPGRADING files updated: [a812d0c](https://github.com/ruby-grape/grape-swagger/pull/922/commits/a812d0c481449a60505eccc5278b15a5674259db).

LeFnord (approved):

LeFnord:
thanks @numbata

PR #924, opened 2024-04-21
fix: Use mount_path to narrow down urls_for

chibicco:
## Overview

Hello, thanks for this useful library !

When you run `rake oapi:fetch`, All endpoints that include the path name doc will be caught.
Narrow down the target path using mount_path.

## Related items

- https://github.com/ruby-grape/grape-swagger/pull/877

### DIFF
--- a/lib/grape-swagger/rake/oapi_tasks.rb
+++ b/lib/grape-swagger/rake/oapi_tasks.rb
@@ -95,7 +95,7 @@ module GrapeSwagger
       def urls_for(api_class)
         api_class.routes
                  .map(&:path)
-                 .select { |e| e.include?('doc') }
+                 .grep(/#{GrapeSwagger::DocMethods.class_variable_get(:@@mount_path)}/)
                  .reject { |e| e.include?(':name') }
                  .map { |e| format_path(e) }
                  .map { |e| [e, ENV.fetch('resource', nil)].join('/').chomp('/') }
--- a/spec/lib/oapi_tasks_spec.rb
+++ b/spec/lib/oapi_tasks_spec.rb
@@ -14,6 +14,10 @@ RSpec.describe GrapeSwagger::Rake::OapiTasks do
       namespace :otherItem do
         get '/'
       end
+
+      namespace :my_doc do
+        get '/'
+      end
     end
 
     class Base < Grape::API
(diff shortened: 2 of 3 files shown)


spec/lib/oapi_tasks_spec.rb
```diff
@@ -115,7 +119,7 @@ class Base < Grape::API
         end
 
         it 'returns complete doc' do
-          expect(response['paths'].length).to eql 2
+          expect(response['paths'].length).to eql 3
```
chibicco, line 122: We added my_doc to check that paths containing "doc" are not targeted by urls_for.
Consequently, the number of tests here has also changed.

spec/support/mock_parser.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'ostruct'
```
chibicco, line 3: The change isn't directly related to the fix, but it seems to be causing some tests to fail.

see: https://github.com/ruby-grape/grape-swagger/actions/runs/8771833802/job/24070096127?pr=924

spec/lib/oapi_tasks_spec.rb
```diff
@@ -132,6 +136,15 @@ class Base < Grape::API
     end
   end
 
+  describe '#urls_for' do
+    require 'pry'
```
LeFnord, line 140: please, remove it

LeFnord (commented):
thanks @chibicco … please rebase against master and add a Changelog entry

chibicco:
I rebased it on master : )

LeFnord (approved):
thanks @chibicco

lib/grape-swagger/rake/oapi_tasks.rb
```diff
@@ -95,7 +95,7 @@ def make_request(url)
       def urls_for(api_class)
         api_class.routes
                  .map(&:path)
-                 .select { |e| e.include?('doc') }
```
diegodorado, line 98: This breaks the specs generation

LeFnord: mmh, it is gone … what do you mean?

fidel-hsiung: Ah, thanks! We’ve fixed it. This explains why my `resource :documents` was always having issues. Upgrading the gem version should solve it.

PR #926, opened 2024-04-26
Refactor route and namespace combination logic

numbata:
## Motivation

The current implementation of route and namespace combinations uses `@target_class.combined_*` accessors that inadvertently trigger hash copies due to `transform_values`. This leads to inefficiencies and unexpected behavior where changes to supposedly original hashes do not persist.

### Shortly

The proposed refactor avoids using `@target_class.combined_*` until the final step, instead using local variables to accumulate and combine routes and namespaces directly. This change prevents premature hash copying and streamlines the combination process, ensuring that changes are properly reflected and improving overall performance.

### TLTR;

All `@target_class.combined_*` calls go through the [`add_setup` method](https://github.com/ruby-grape/grape/blob/v2.0.0/lib/grape/api.rb#L146) of the `grape` gem (caused by [`override_all_methods!`](https://github.com/ruby-grape/grape/blob/v2.0.0/lib/grape/api.rb#L51-L58)). In the depths of the `add_setup` method [it does](https://github.com/ruby-grape/grape/blob/v2.0.0/lib/grape/api.rb#L183):

```ruby
argument.transform_values { |value| evaluate_arguments(configuration, value).first }
```

The `transform_values` makes a copy of the original hash object and returns it. This leads to unexpected behavior in cases such as:

```ruby
@target_class.combined_routes[resource] ||= []
```

Where expected flow:

1. The `@target_class.combined_routes` returns an original hash.
2. The empty array sets under the `resource` key on the original hash.
3. The original `combined_routes` is filled with a new key-value.

But actual flow:

1. The `@target_class.combined_routes` creates a copy of the empty hash
2. The empty array is set to the empty hash under the `resource` key
3. The original `combined_routes` is still an empty hash.

### DIFF
--- a/lib/grape-swagger.rb
+++ b/lib/grape-swagger.rb
@@ -24,7 +24,7 @@ module SwaggerRouting
   private
 
   def combine_routes(app, doc_klass)
-    app.routes.each do |route|
+    app.routes.each_with_object({}) do |route, combined_routes|
       route_path = route.path
       route_match = route_path.split(/^.*?#{route.prefix}/).last
       next unless route_match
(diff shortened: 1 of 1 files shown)


numbata:
@LeFnord just a friendly reminder, no rush.

LeFnord:
no problem … I'll add the entry by the next release

PR #927, opened 2024-05-07
Set default parameter location based on consumes

dhruvCW:
Just a fork of #880 that is based on the latest changes in master.


.rubocop.yml
```diff
@@ -75,3 +75,6 @@ Style/RegexpLiteral:
 
 Style/SlicingWithRange:
   Enabled: false
+
+Metrics/ParameterLists:
```
LeFnord, line 79: please remove it here, and add `# rubucop:disable Metrics/ParameterLists` before the method and after `# rubucop:enable Metrics/ParameterLists`

lib/grape-swagger/doc_methods/parse_params.rb
```diff
@@ -148,14 +148,18 @@ def document_example(settings)
...
-        def param_type(value_type)
+        def param_type(value_type, consumes)
           param_type = value_type[:param_type] || value_type[:in]
           if value_type[:path].include?("{#{value_type[:param_name]}}")
             'path'
           elsif param_type
             param_type
           elsif %w[POST PUT PATCH].include?(value_type[:method])
-            DataType.request_primitive?(value_type[:data_type]) ? 'formData' : 'body'
+            if consumes.include?('application/x-www-form-urlencoded') || consumes.include?('multipart/form-data')
```
LeFnord, line 158: what is if `consumes` isn't defined?

LeFnord (commented):
thanks @dhruvCW …
but I have one big question, ite seems now the consumes key must be given,
is ot what you want? 
want to see a fallback, if it isn't defined then json should be consumed

lib/grape-swagger/doc_methods/parse_params.rb
dhruvCW: the method expects `consumes` to be an array of consumed types i.e. `application/json` or `application/x-www-form-urlencoded`.

consumes is defined by [on this line](https://github.com/ruby-grape/grape-swagger/blob/master/lib/grape-swagger/endpoint.rb#L121) and by default (if consumes isn't defined) is set to `application/json`.

This PR basically changes the default value of `in` field to `body` unless the consumes value is explicitly set to `urlencoded`.

Would you like me to define a default value for consumes in the `call` method as well ?

LeFnord (approved):
yeah @dhruvCW it makes sense, thanks

PR #929, opened 2024-05-11
Set query parameter for array of primitive types

### LINKED ISSUE #883: Does not support query array  parameters
Currently, `grape-swagger` does not support query array parameters, setting them to `formData` by default:

https://github.com/ruby-grape/grape-swagger/blob/master/lib/grape-swagger/doc_methods/parse_params.rb#L87

```
            param_type = value_type[:documentation][:param_type]
            doc_type = value_type[:documentation][:type]
            type = DataType.mapping(doc_type) if doc_type && !DataType.request_primitive?(doc_type)
            collection_format = value_type[:documentation][:collectionFormat]
          end

          param_type ||= value_type[:param_type]

          array_items = parse_array_item(
          array_items = parse_array_item(
            definitions,
            type,
            value_type
          )

          @parsed_param[:in] = param_type || 'formData'
          @parsed_param[:items] = array_items
```

This situations where as long as it is an array parameter, even if it occurs in a `get` request, by default the parameter's `in` is set to `formData`.

However, according to the [OpenAPI v2 spec](https://swagger.io/docs/specification/2-0/describing-parameters/#query-parameters), `Query parameters only support primitive types. You can have an array, but the items must be a primitive value type. Objects are not supported.`

As such, in situations where there an array parameter's `in`/`param_type` is not explicitly defined and it occurs in a `GET` request and the array items type are non-object primitive value types, it should be set as `query` type.

dhruvCW:
Just a rebase of #886  onto the latest changes in master.

Fixes #883

### DIFF
--- a/lib/grape-swagger/doc_methods/parse_params.rb
+++ b/lib/grape-swagger/doc_methods/parse_params.rb
@@ -21,7 +21,7 @@ module GrapeSwagger
           # optional properties
           document_description(settings)
           document_type_and_format(settings, data_type)
-          document_array_param(value_type, definitions) if value_type[:is_array]
+          document_array_param(value_type, definitions, consumes) if value_type[:is_array]
           document_default_value(settings) unless value_type[:is_array]
           document_range_values(settings) unless value_type[:is_array]
           document_required(settings)
@@ -68,15 +68,12 @@ module GrapeSwagger
           GrapeSwagger::DocMethods::Extensions.add_extensions_to_root(settings, @parsed_param)
         end
 
-        def document_array_param(value_type, definitions)
+        def document_array_param(value_type, definitions, consumes)
           if value_type[:documentation].present?
-            param_type = value_type[:documentation][:param_type] || value_type[:documentation][:in]
             doc_type = value_type[:documentation][:type]
             type = DataType.mapping(doc_type) if doc_type && !DataType.request_primitive?(doc_type)
             collection_format = value_type[:documentation][:collectionFormat]
           end
-
-          param_type ||= value_type[:param_type]
 
           array_items = parse_array_item(
             definitions,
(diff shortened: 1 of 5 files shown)


dhruvCW:
Failing tests seem to be because of a network hiccup 😅

lib/grape-swagger/doc_methods/parse_params.rb
```diff
@@ -21,7 +21,7 @@ def call(param, settings, path, route, definitions, consumes) # rubocop:disable
           # optional properties
           document_description(settings)
           document_type_and_format(settings, data_type)
-          document_array_param(value_type, definitions) if value_type[:is_array]
+          document_array_param(value_type, definitions, consumes) if value_type[:is_array]
```
LeFnord, line 24: 🤔 … is it related to this PR, thought it was introduced #927

LeFnord (commented):
thanks @dhruvCW … is it an addition to #927?
if yes, then I think some things can be refactored, especially stuff from the `array_param_type` method, or is it an rebase mistake?

lib/grape-swagger/doc_methods/parse_params.rb
dhruvCW: it is related to this PR (handling for `in[:query]` for get requests but it is indeed something that was missed in #927

dhruvCW:
> thanks @dhruvCW … is it an addition to #927? if yes, then I think some things can be refactored, especially stuff from the `array_param_type` method, or is it an rebase mistake?

it indeed builds on top of #927 specifically handling the case when an array parameter is incorrectly mapped as `formData` even when it's part of a `GET` request. This PR also fixes the problem wherein `array` parameters always default to `formData` (this was missed in #927)

dhruvCW:
> I think some things can be refactored, especially stuff from the array_param_type method, or is it an rebase mistake?

could you point them out please 😅 I just rebased the change so I haven't given the implementation itself much thought 😅

LeFnord (approved):

LeFnord:
thanks @dhruvCW

dhruvCW:
thanks for merging the change ❤️ . Could we get a release soon ? This and the previous PR would help fix a long standing issue with the generated swagger docs.

LeFnord:
released

PR #931, opened 2024-05-17
Use `route.namespace` instead instance_variable_get(@options)

ericproulx:
While trying to improve some parts of Grape, I came across an error in your gem since my code had removed `@options` in `Grape::Router::Route`. This PR replaces the `instance_variable_get` by simply using `route.namespace` which is accessible from an instance of [Grape::Router::Route](https://github.com/ruby-grape/grape/blob/82f1a097106b3a717f8053fe820481d04af061c8/lib/grape/router/route.rb#L13).

I also refactor some small methods like `route_namespace_equals?` by using an Enumerator instead of `if` with interpolation. Ain't a big change but it will lazily interporlate strings if needed.

### DIFF
--- a/lib/grape-swagger.rb
+++ b/lib/grape-swagger.rb
@@ -45,12 +45,10 @@ module SwaggerRouting
   end
 
   def determine_namespaced_routes(name, parent_route, routes)
-    if parent_route.nil?
-      routes.values.flatten
-    else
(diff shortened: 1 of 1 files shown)


LeFnord:
thanks @ericproulx … but it's not my gem 😉

PR #934, opened 2024-06-22
Document the min/max Items/Length fields if the attribute uses the length validator

dhruvCW:
Grape 2.1.0 Introduced the [length validator ](https://github.com/ruby-grape/grape/pull/2437). We can now use the length validator to specify the `min/maxItems` (for array types) and `min/maxLength` (for String types) for the attributes that use the validator in the generated swagger document.

Fixed test that was broken with rack 3 and restricted `rack < 3` for `grape < 2`

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -19,7 +19,13 @@ group :development, :test do
   gem 'pry', platforms: [:mri]
   gem 'pry-byebug', platforms: [:mri]
 
-  gem 'rack'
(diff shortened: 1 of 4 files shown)


dhruvCW:
Test failures seem unrelated to this PR 🤔

LeFnord (approved):
thanks @dhruvCW … yeah it makes sense

PR #940, opened 2024-09-16
Grape 2.2.0 compatibility

### LINKED ISSUE #939: Compatibility with grape 2.2.0
Running on the latest 2.2.0 release of grape will produce these exceptions:

    NameError:
      uninitialized constant Grape::ContentTypes::CONTENT_TYPES
    # ./lib/grape-swagger/endpoint.rb:15:in `content_types_for'

    NoMethodError:
      undefined method `formatters' for module Grape::Formatter
    # ./lib/grape-swagger/endpoint.rb:14:in `content_types_for'

Used here:
* https://github.com/ruby-grape/grape-swagger/blob/v2.1.0/lib/grape-swagger/doc_methods/produces_consumes.rb#L10
* https://github.com/ruby-grape/grape-swagger/blob/v2.1.0/lib/grape-swagger/endpoint.rb#L15

Grape has removed/renamed these constants/methods via https://github.com/ruby-grape/grape/commit/fb67ea9940ea47d353cc9b413a7e25fc8da0983e#diff-e28e889e3579545f8e2fedc3ab52f78530206ed808afdc9a23703f8b089f80e3

padde:
Running on grape 2.0.2 resulted in the following exceptions:

    NameError:
      uninitialized constant Grape::ContentTypes::CONTENT_TYPES
    # ./lib/grape-swagger/endpoint.rb:15:in `content_types_for'

    NoMethodError:
      undefined method `formatters' for module Grape::Formatter
    # ./lib/grape-swagger/endpoint.rb:14:in `content_types_for'

This is due to internal refactorings in grape as of v2.2.0, specifically https://github.com/ruby-grape/grape/commit/fb67ea99

Fixes #939

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -119,6 +119,70 @@ jobs:
         bundle update
         bundle exec rspec
 
+  grape-21:
+    runs-on: ubuntu-latest
+    needs: ['rubocop']
+    env:
+      GRAPE_VERSION: '2.1.3'
+    strategy:
+      matrix:
+        ruby-version: ['3.0', '3.1', '3.2', '3.3', 'head']
+    steps:
+    - name: Check out branch
+      uses: actions/checkout@v4
+    - name: Set up Ruby
+      uses: ruby/setup-ruby@v1
+      with:
+        ruby-version: ${{ matrix.ruby-version }}
+    - name: Run rspec wo model parser
+      run: |
+        bundle update
+        bundle exec rspec
+    - name: Run rspec w entity parser
+      env:
+        MODEL_PARSER: grape-swagger-entity
+      run: |
+        bundle update
+        bundle exec rspec
+    - name: Run rspec w representable parser
+      env:
+        MODEL_PARSER: grape-swagger-representable
+      run: |
+        bundle update
+        bundle exec rspec
+
+  grape-22:
+    runs-on: ubuntu-latest
+    needs: ['rubocop']
+    env:
+      GRAPE_VERSION: '2.2.0'
+    strategy:
+      matrix:
+        ruby-version: ['3.0', '3.1', '3.2', '3.3', 'head']
+    steps:
+    - name: Check out branch
+      uses: actions/checkout@v4
+    - name: Set up Ruby
+      uses: ruby/setup-ruby@v1
+      with:
+        ruby-version: ${{ matrix.ruby-version }}
+    - name: Run rspec wo model parser
+      run: |
+        bundle update
+        bundle exec rspec
+    - name: Run rspec w entity parser
+      env:
+        MODEL_PARSER: grape-swagger-entity
+      run: |
+        bundle update
+        bundle exec rspec
+    - name: Run rspec w representable parser
+      env:
+        MODEL_PARSER: grape-swagger-representable
+      run: |
+        bundle update
+        bundle exec rspec
+
   grape-HEAD:
     runs-on: ubuntu-latest
     needs: ['rubocop']
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,182 +22,52 @@ jobs:
           bundler-cache: true
       - name: Run rubocop
         run: bundle exec rubocop --parallel --format progress
-
-  grape-17:
+  test:
+    strategy:
+      matrix:
+        entry:
+          - { ruby: '3.0', grape: '1.7.1' }
+          - { ruby: '3.1', grape: '1.7.1' }
+          - { ruby: '3.2', grape: '1.7.1' }
+          - { ruby: '3.3', grape: '1.7.1' }
+          - { ruby: 'head', grape: '1.7.1' }
+          - { ruby: '3.0', grape: '1.8.0' }
+          - { ruby: '3.1', grape: '1.8.0' }
+          - { ruby: '3.2', grape: '1.8.0' }
+          - { ruby: '3.3', grape: '1.8.0' }
+          - { ruby: 'head', grape: '1.8.0' }
+          - { ruby: '3.0', grape: '2.0.0' }
+          - { ruby: '3.1', grape: '2.0.0' }
+          - { ruby: '3.2', grape: '2.0.0' }
+          - { ruby: '3.3', grape: '2.0.0' }
+          - { ruby: 'head', grape: '2.0.0' }
+          - { ruby: '3.0', grape: '2.1.3' }
+          - { ruby: '3.1', grape: '2.1.3' }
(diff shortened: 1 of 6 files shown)


grape-swagger.gemspec
```diff
@@ -15,7 +15,7 @@ Gem::Specification.new do |s|
   s.metadata['rubygems_mfa_required'] = 'true'
 
   s.required_ruby_version = '>= 3.0'
-  s.add_runtime_dependency 'grape', '>= 1.7', '< 3.0'
+  s.add_runtime_dependency 'grape', '>= 2.2.0', '< 3.0'
```
padde, line 18: Not sure if it's OK to drop support for older grape versions - please let me know.

Otherwise we'd probably need to introduce dynamic behaviour at runtime based on the currently loaded grape version, something like

```ruby
if Gem::Version.new(Grape::VERSION) < '2.0.2'
  # old behaviour
else
  # new behaviour
end
```

README.md
```diff
@@ -54,6 +54,7 @@ The following versions of grape, grape-entity and grape-swagger can currently be
 | 0.34.0        | 2.0          | >= 0.16.2 ... < 1.3.0   | >= 0.5.0     | >= 2.4.1      |
 | >= 1.0.0      | 2.0          | >= 1.3.0                | >= 0.5.0     | >= 2.4.1      |
 | >= 2.0.0      | 2.0          | >= 1.7.0                | >= 0.5.0     | >= 2.4.1      |
+| >= 3.0.0      | 2.0          | >= 2.2.0                | >= 0.5.0     | >= 2.4.1      |
```
padde, line 57: If the change is made backwards-incompatible as suggested, a major release must follow.

grape-swagger.gemspec
```diff
@@ -15,8 +15,8 @@ Gem::Specification.new do |s|
   s.metadata['rubygems_mfa_required'] = 'true'
 
   s.required_ruby_version = '>= 3.0'
-  s.add_runtime_dependency 'grape', '>= 1.7', '< 3.0'
-  s.add_runtime_dependency 'rack-test', '~> 2'
+  s.add_dependency 'grape', '>= 2.2.0', '< 3.0'
```
padde, line 18: Not sure if it's OK to drop support for older grape versions - please let me know.

Otherwise we'd probably need to introduce dynamic behaviour at runtime based on the currently loaded grape version, something like

```ruby
if Gem::Version.new(Grape::VERSION) < '2.2.0'
  # old behaviour
else
  # new behaviour
end
```

grape-swagger.gemspec
```diff
@@ -15,8 +15,8 @@ Gem::Specification.new do |s|
...
+  s.add_dependency 'rack-test', '~> 2'
```
padde, line 19: The switch from `add_runtime_dependency` to `add_dependency` was made to satisfy Rubocop.

.github/workflows/ci.yml
```diff
@@ -91,7 +91,7 @@ jobs:
     runs-on: ubuntu-latest
     needs: ['rubocop']
     env:
-      GRAPE_VERSION: '2.0.0'
+      GRAPE_VERSION: '2.2.0'
```
dblock, line 94: I think this is where we expand the matrix to test against multiple versions of Grape, 2.0 and 2.2.

README.md
dblock: Let's try not to make a backwards incompatible one and break things only if we have to?

grape-swagger.gemspec
dblock: I would prefer we didn't unless we absolutely have to.

dblock (changes requested):
Instead of relying on a Grape constant that's not supposed to be part of the public interface, just copy it into grape-swagger?

.github/workflows/ci.yml
padde: What about 2.1, should we test that as well?

padde:
> Instead of relying on a Grape constant that's not supposed to be part of the public interface, just copy it into grape-swagger? Expand CI to use multiple versions of Grape to make sure our compatibility is tested well.

@dblock sure, we could do that and it would vastly simplify the implementation, as we wouldn't have to massage the internal constants as much to make it work for our purposes here.

padde:
@dblock looking further into the code that is used from grape here, it seems that there always has been a flaw in the way we retrieve content types in grape-swagger. Grape [allows adding user-defined content types](https://github.com/ruby-grape/grape?tab=readme-ov-file#api-formats). However, if I get this right, grape-swagger never even considered the ones coming from the settings, just the default values form the internal constants that are now causing the issue.

.github/workflows/ci.yml
padde: I've added both 2.1 and 2.2 to CI for now and removed the initial attempt to fix the issues, let's talk about a way to solve this.

dblock (changes requested):
The way you implemented this create a lot of repetition for the variable. Can you try and please turn this into a matrix with two entries: a ruby version and a grape version? We don't need to test all possible permutations either. Here's an example of one: https://github.com/mongoid/mongoid-locker/blob/master/.github/workflows/test.yml

padde:
> Can you try and please turn this into a matrix with two entries: a ruby version and a grape version?

@dblock sure. Honestly, I just expanded on the existing approach to avoid this discussion but I am happily reducing the duplication!

> We don't need to test all possible permutations either.

That is a bit vague - which permutations should we test then, if not all? I will for now follow the "wasteful" approach with all permutations, feel free to tell me which ones to remove, if you like.

.github/workflows/ci.yml
```diff
@@ -22,182 +22,52 @@ jobs:
...
+          - { ruby: '3.0', grape: '2.2.0' }
+          - { ruby: '3.1', grape: '2.2.0' }
+          - { ruby: '3.2', grape: '2.2.0' }
+          - { ruby: '3.3', grape: '2.2.0' }
+          - { ruby: 'head', grape: '2.2.0' }
+          - { ruby: '3.0', grape: 'HEAD' }
+          - { ruby: '3.1', grape: 'HEAD' }
+          - { ruby: '3.2', grape: 'HEAD' }
+          - { ruby: '3.3', grape: 'HEAD' }
+          - { ruby: 'head', grape: 'HEAD' }
```
padde, line 58: @dblock let me know if you want any of these combinations removed, I am unsure about which ones you want to be "officially unsupported/untested".

dblock: As long as they pass I'm good with them! Computers work for us.

dblock (approved):

Issue #325, opened 2015-12-29
Add example to path param


herberzt:
I have been adding examples to some of my parameters, but up to this point they have not been path params. I tried to do it with the following but it doesn't show up. 
```route_param :id, desc: "User ID", documentation: { example: User.first.id }  do```

mneumark:
I can't even get a description to show up for a `route_param`. +1 Must fix.

dblock:
Yeah these aren't first class `params` in Grape. It needs work.

thimios:
:+1:

rjaus:
I'm facing this issue as well. I've written two tests that related to documentation & examples for route params.

https://github.com/rjaus/grape-swagger/commit/dd98a217b166ca3799eb4fff8fae335a67202a3a

I'm happy to try and fix it too, if someone will point me in the right direction. @dblock where does this need to be fixed? What needs to be changed?

dblock:
> I'm happy to try and fix it too, if someone will point me in the right direction. @dblock where does this need to be fixed? What needs to be changed?

in this project, but I don't know specifically where

dblock:
Can someone please verify whether https://github.com/ruby-grape/grape-swagger/pull/943 fixes this?

PR #943, opened 2024-10-12
Fix route_param documentation and type

### LINKED ISSUE #942: Type, description and documentation is ignored when mounting class inside route_param
Example:

```ruby
class MyApi < Grape::API
  route_param :account_id, type: String, desc: 'id of account', documentation: { desc: 'another-desc' } do
    mount AccountsApi

    get '/another-list' do
      []
    end
  end

  add_swagger_documentation
end

class AccountsApi < Grape::API
  get '/list' do
    []
  end
end
```

Resulting documentation:

```json
"/{account_id}/list": {
  "get": {
    "parameters": [
      {
        "in": "path",
        "name": "account_id",
        "type": "integer",
        "format": "int32",
        "required": true
      }
    ],
  }
},
"/{account_id}/another-list": {
  "get": {
    "parameters": [
      {
        "in": "path",
        "name": "account_id",
        "type": "string",
        "required": true
      }
    ],
  }
}
```

- For `account_id` param of `/list` type is incorrect, description and documentation is ignored
- For `account_id` param of `/another-list` type is correct, but description and documentation is still ignored

4ndv:
`Grape::Entity#merge_params` returns `route.params`, which effectively discards result of code above this line, because `route.params` produces new object each time it's called

Fixes #942 and (potentially) #325

Side note: I think filter condition in `delete_if` at least requires an explanation comment/revision, because it's a little bit cryptic

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 #### Fixes
 
+* [#943](https://github.com/ruby-grape/grape-swagger/pull/943): Fix route_param documentation and type - [@4ndv](https://github.com/4ndv)
 * Your contribution here.
 
 
--- a/lib/grape-swagger/endpoint.rb
+++ b/lib/grape-swagger/endpoint.rb
@@ -377,7 +377,7 @@ module Grape
         route_params[key] = path.merge(params)
       end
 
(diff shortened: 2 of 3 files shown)


dblock:
Looks right, thanks for fixing this!

PR #944, opened 2024-10-22
Amend typoes

pieterocp:
Saw a couple of typographic errors when reading the readme, so addressed them + ran some IDE tooling to sniff out some of the others.


dblock (approved):
Thanks!

PR #957, opened 2025-05-21
Fix issue where array params are sometimes omitted when array_use_braces is true

### LINKED ISSUE #952: Bug report: omitted array parameters with `array_use_braces: true`
When `array_use_braces` is set to `true`, array parameters are omitted when they are body parameters but don't explicitly specify `documentation: { param_type: "body" }`.

Example:
```ruby
# with add_swagger_documentation(array_use_braces: true)

params do
  requires :param1, type: Array[String]
  requires :param2, type: Array[Integer], documentation: { param_type: "body" }
end
post :foo do
  { declared_params: declared(params)
end
```

The spec will look like:
```json
"postFoo": {
  "required": ["param2"],
  "type": "object",
  "properties": {
    "param2": {
      "type": "array",
      "items": { "type": "integer", "format": "int32" }
    }
  }
}
```

It is particularly annoying if one wants to share some params between GET and POST/PUT/PATCH endpoints and rely on the implicit behavior to have the params be query params or body params as appropriate.

olivier-thatch:
Fixes #952.

The fix consists of removing the `[]` suffix at the end of a parameter's name if the parameter is an array and it's a body param.

This fix is very hacky: ideally, `[]` shouldn't be added in the first place. But I don't think that can be achieved without significantly refactoring the code.

@dblock @numbata Curious if you have any thoughts on this approach.

### DIFF
--- a/lib/grape-swagger/doc_methods/parse_params.rb
+++ b/lib/grape-swagger/doc_methods/parse_params.rb
@@ -17,6 +17,7 @@ module GrapeSwagger
             in: param_type(value_type, consumes),
             name: settings[:full_name] || param
           }
+          fix_name_of_body_array_param(value_type)
 
           # optional properties
           document_description(settings)
@@ -34,6 +35,12 @@ module GrapeSwagger
         end
 
         private
+
+        def fix_name_of_body_array_param(value_type)
+          return unless @parsed_param[:in] == 'body' && value_type[:is_array]
+
+          @parsed_param[:name] = @parsed_param[:name].delete_suffix('[]')
+        end
 
         def document_description(settings)
           description = settings[:desc] || settings[:description]
--- a/spec/swagger_v2/params_array_spec.rb
+++ b/spec/swagger_v2/params_array_spec.rb
@@ -95,6 +95,27 @@ describe 'Group Params as Array' do
             { 'declared_params' => declared(params) }
           end
 
+          helpers do
+            params :common_params do
+              requires :array_of_string, type: Array[String]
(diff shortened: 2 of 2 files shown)


dblock:
CI was fixed in https://github.com/ruby-grape/grape-swagger/pull/958, rebase please?

numbata:
Hi @olivier-thatch, a quick fix for this PR could be:

```diff
diff --git a/lib/grape-swagger/doc_methods/move_params.rb b/lib/grape-swagger/doc_methods/move_params.rb
index c8417ce..12c6477 100644
--- a/lib/grape-swagger/doc_methods/move_params.rb
+++ b/lib/grape-swagger/doc_methods/move_params.rb
@@ -136,7 +136,7 @@ module GrapeSwagger
           return if value.blank?
 
           definition[:required] ||= []
-          definition[:required].push(*value)
+          definition[:required].push(*value).uniq!
         end
 
         def build_body_parameter(name, options)
```

I considered using a `Set.new` for `definition[:required]` to enforce uniqueness, but that would change the interface (as example, `Set` doesn’t support `.push`) and could introduce side effects for other `grape-swagger-*` gems.

numbata:
I looked more closely at the PR and noticed we have incorrect behavior for nested arrays. The type `Array[Array[String]]` collapses into `String`, so `document_array_param` only ever sees one dimension. As a result, a declaration like
`requires :matrix, type: Array[Array[String]]`
produces `{ type: 'array', items: { type: 'string' } }`, which loses the inner array.

We need to either fix this to properly support multi-dimensional arrays or clearly document that they aren’t supported.

PR #959, opened 2025-05-29
Add support for procs in default value

seriousdev-gh:
Grape supports lambda and procs as default values, it evaluates them for each request. But grape-swagger doesn`t evaluate procs, it shows stringified representation of proc, as in screenshot: 

image

This pull request fixes it.

### DIFF
--- a/lib/grape-swagger/doc_methods/parse_params.rb
+++ b/lib/grape-swagger/doc_methods/parse_params.rb
@@ -52,7 +52,9 @@ module GrapeSwagger
         end
 
         def document_default_value(settings)
-          @parsed_param[:default] = settings[:default] if settings.key?(:default)
+          return unless settings.key?(:default)
+
+          @parsed_param[:default] = settings[:default].is_a?(Proc) ? settings[:default].call : settings[:default]
         end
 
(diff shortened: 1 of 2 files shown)


numbata:
Hi @seriousdev-gh!

Thanks for adding Proc support for default values - I really appreciate the added flexibility! But I do have a small concern: executing arbitrary Procs at definition time could introduce unexpected performance or side-effect risks. For example, someone could write a Proc that:

 - Makes an HTTP call with no timeout.
 - Runs a synchronous DB query that sleeps for several seconds.
 - Writes to a log file.
 - Depends on something non-obvious.

Additionally, grape-swagger invokes the Proc outside of the usual Grape request context (see how `DefaultValidator` works [here](https://github.com/ruby-grape/grape/blob/master/lib/grape/validations/validators/default_validator.rb#L17)), which might lead to subtle differences in behavior or potential errors even.

Would it make sense to avoid immediately calling the Proc and provide some placeholder instead? For example, we can check do we have something non-proc-able in documentation section? Or simply omit the `default` field for Procs.

This way we preserve Proc support for real requests, but avoid any risks during doc generation.

PR #978, opened 2026-05-17
Fix Grape 3.2+ compatibility: desc kwargs, custom types, multi-type param recovery

### LINKED ISSUE #977: Grape deprecates passing a positional options Hash to `desc` (grape-swagger uses it internally)
### Summary

Grape ([ruby-grape/grape#2723](https://github.com/ruby-grape/grape/pull/2723)) is moving `Grape::DSL::Desc#desc` to **keyword arguments**. Passing a positional options Hash still works but now emits a deprecation:

> Passing a positional options Hash to `desc` is deprecated. Pass keyword arguments instead.

For apps that configure `Grape.deprecator.behavior = :raise` (the default in Grape's own test suite, and common in CI), this turns into an `ActiveSupport::DeprecationException` and `add_swagger_documentation` raises outright.

### Where grape-swagger triggers it

grape-swagger 2.1.4 calls `desc` with a positional Hash in `lib/grape-swagger/doc_methods.rb`:

```ruby
# ~line 96
desc api_doc.delete(:desc), api_doc

# ~line 108
desc specific_api_doc.delete(:desc), { params: specific_api_doc.delete(:params) || {}, **specific_api_doc }
```

### Suggested fix

Double-splat the options Hash so they are passed as keyword arguments (backward compatible with current Grape):

```ruby
desc api_doc.delete(:desc), **api_doc

desc specific_api_doc.delete(:desc), params: specific_api_doc.delete(:params) || {}, **specific_api_doc
```

### Reproduction

```ruby
require 'grape'
require 'grape-swagger'

Grape.deprecator.behavior = :raise

Class.new(Grape::API) { add_swagger_documentation }
# => ActiveSupport::DeprecationException: Passing a positional options Hash to `desc` is deprecated.
```

Tested against grape-swagger 2.1.4 with Grape from [ruby-grape/grape#2723](https://github.com/ruby-grape/grape/pull/2723). That PR adds a grape-swagger integration job to Grape's CI to catch regressions like this going forward.

numbata:
## Summary

This PR brings grape-swagger into full compatibility with Grape 3.2 and beyond. It fixes two breaking changes introduced in Grape 3.2+ — the `desc` keyword-argument deprecation (which raises when `Grape.deprecator.behavior = :raise`) and the rejection of string type names and non-`.parse` custom types in `params` blocks — and adds recovery of `type: [A, B]` multi-type params, which Grape 3.2+ was serialising as a raw coercer string in swagger output instead of the declared type. The minimum Grape dependency is bumped to `>= 2.1` (Grape 1.8 and 2.0 were already failing on Ruby 3.3+ due to an upstream Mustermann incompatibility), and the upper bound is expanded to `< 5.0` to support Grape 4.x.

### `lib/grape-swagger/doc_methods.rb`
- Pass keyword arguments to both `desc` calls so Grape's deprecator does not fire (or raise when `behavior = :raise`).
- Accept string-keyed `api_documentation` / `specific_api_documentation` hashes (e.g. loaded from YAML/JSON) via `transform_keys(&:to_sym)`.
- Accept `:description` as an alias for `:desc`; explicit `desc: nil` wins and does not fall through.

### `lib/grape-swagger/request_param_parsers/route.rb`
- Add `collect_variant_types` + `extract_variant_validator_parts` + `restore_variant_type` to recover the actual type list from `VariantCollectionCoercer` when Grape serialises `type: [A, B]` via `#to_s`. Grape 3.2+ stores validator metadata as Hash entries; older supported versions use `CoerceValidator` object instances — both shapes are handled.

### `spec/support/model_parsers/mock_parser.rb` and `representable_parser.rb`
- Add `def self.parse(val) = val` to `NestedModule::ApiResponse`. Grape 3.2+ requires unknown types to satisfy `Types.custom?` (respond to `.parse` with arity 1); `Grape::Entity` already provides it so `entity_parser.rb` needs no change.

### `spec/swagger_v2/params_example_spec.rb`
- Move `type: 'Object'` from the Grape `params` declaration into `documentation: { type: 'Object' }`. Grape 3.2+ rejects string type names in params blocks; grape-swagger picks the type from the merged settings hash so swagger output is unchanged.

### `grape-swagger.gemspec`
- Bump minimum Grape to `>= 2.1` (was `>= 1.7`); expand upper bound to `< 5.0` (was `< 4.0`) to support Grape 4.x HEAD.

### `.github/workflows/ci.yml`
- Drop Grape 1.8.0 and 2.0.0 matrix rows (failing on Ruby 3.3+ upstream).
- Add Ruby 3.2–3.4 × Grape 3.2.1 rows.
- Drop `ruby: '3.1', grape: '3.2.1'` (Grape 3.2.1 requires Ruby >= 3.2).

### `.rubocop.yml`
- Disable `Style/OneClassPerFile` — `lib/grape-swagger.rb` defines `SwaggerRouting` and `SwaggerDocumentationAdder` as separate top-level modules; moving them into the `GrapeSwagger` namespace would be a breaking change and will ship separately.

### `Gemfile`
- Add `multi_json` explicitly — required transitively by `representable` but not bundled on Ruby HEAD.

## Validation

| MODEL_PARSER | Examples | Failures |
|---|---|---|
| mock (default) | 493 | 0 |
| grape-swagger-entity | 493 | 0 |
| grape-swagger-representable | 493 | 0 |

Tested against Grape 2.1.3, 2.2.0, 3.2.1, and HEAD.

Closes #977


dblock (approved):

numbata:
@dblock FYI, I’ll jump back into this and the other PRs in the my grape-review queue later today. It looks like we have everything we need to release the new minor version of the gem.

dblock:
CI does need to be fixed

dblock:
Since this breaks backwards compat, needs a version bump and an UPGRADING note. Check compatibility in README, too.

moskvin (approved):

numbata:
@moskvin Thanks for taking a look. I’ll combine this PR with yours for the next release. But as @dblock mentioned, there are still a few changes that should be made before merging, and I completely agree with that.

dblock (approved):
LGTM, merge at will

PR #980, opened 2026-06-25
Refactor swagger documentation modules into GrapeSwagger namespace with routing fixes

numbata:
## Summary

This PR builds on top of [#976](https://github.com/ruby-grape/grape-swagger/pull/976) by [@moskvin](https://github.com/moskvin), who deserves primary credit for the work here. The core refactor — extracting `SwaggerRouting` and `SwaggerDocumentationAdder` out of the top-level namespace into `GrapeSwagger::`, the sibling-namespace prefix boundary fix, the regex injection hardening, and the path-param deep-merge rewrite — was all done by @moskvin. This PR carries that work forward with additional spec coverage, edge-case fixes, and cleanup that emerged from review.

Specific additions on top of #976: remove the conditional guard that could leave `SwaggerDocumentationAdder` unaliased if `SwaggerRouting` was pre-defined elsewhere; fix empty-string prefix generating `//`-prefixed route patterns; add regression specs pinning the `delete_prefix` colon-stripping contract, the symbol-key lookup chain, two-level-deep hash merge, and intentional array-replace behavior in `merge_path_param_options`; remove redundant inline comments that narrated obvious code.

## Changes

- `SwaggerRouting` and `SwaggerDocumentationAdder` extracted to `lib/grape-swagger/swagger_routing.rb` and `lib/grape-swagger/swagger_documentation_adder.rb` under the `GrapeSwagger::` namespace; top-level constants kept as deprecated aliases for backward compatibility
- `Regexp.escape` applied to `route.prefix` and `doc_klass.mount_path` in `combine_routes` to close regex injection
- Sibling-namespace prefix boundary fix: `start_with?(ns_name)` replaced with `name == ns_name || name.start_with?("#{ns_name}/")` so `users` is no longer a phantom child of `user`
- `standalone_namespaces` hoisted out of the per-namespace loop (was O(n) recomputation)
- Empty-string prefix treated as no prefix in `route_path_start_with?` to avoid `//name` patterns
- `gsub(/\/{2,}/, '/')` replaces `sub` so all repeated slash runs in `full_namespace` are normalized, not just the first
- Path-param merging in `request_param_parsers/route.rb` rewritten to deep-merge inherited namespace options and normalize keys to symbols
- Compatibility aliases always assigned unconditionally (removed `unless defined?` guard)
- Spec classes `Spec` renamed to `Issue537Spec` / `Issue579Spec` to avoid cross-file constant leaks

## Related

Builds on #976 by @moskvin.


numbata:
Hey @moskvin — this PR builds directly on top of your work in #976. Would love your eyes on it since you're the original author of these changes. Let us know if anything looks off.

PR #983, opened 2026-07-06
Read route metadata via readers instead of route.options[]

ericproulx:
Consume route metadata through the route's public reader methods instead of reaching into `route.options[...]`.

### What changed

Every `route.options[...]` read in `lib/` is replaced with the equivalent reader:

- `deprecated`, `security`, `summary`, `detail`, `produces`, `is_array`, `hidden`, `nickname`, `tags` → their readers
- `success` / `failure` (keyword aliases of `entity` / `http_codes`) → `route.success` / `route.failure`
- `body_name` → resolved by the caller as `route.body_name`; `build_body_parameter` now takes the resolved value
- the dynamic producer lookup (`:formats` / `:content_types` / `:produces`) → `route.public_send(producer)`
- grape-swagger's own `:desc` / `:default_response` → `route.desc` / `route.default_response`

There is one remaining route.options[...] read: route.options.key?(:tags) in method_object (see "Behavior note" below) — no reader distinguishes "tags absent" from "tags explicitly nil", so presence has to be checked on the raw options hash

### Why

Grape is moving desc metadata toward first-class readers (as it already did for `params` / `anchor` / `requirements`) and wants to reserve `route.options` for user custom keys. Consuming routes through their method interface means grape-swagger keeps working when Grape restructures the options bag. The readers resolve on the currently-supported Grape (verified on 2.4.0).

### Behavior note (`detail: nil`)

`summary_object` / `description_object` previously gated on `route.options.key?(:detail)`; they now read `route.detail`. The only observable difference is when `detail:` is set **explicitly to `nil`** — it is now treated the same as an absent `detail` (the description holds the text and there is no separate summary), which is the more sensible reading of "no detail". Covered by a new example in `api_swagger_v2_detail_spec.rb`.

### Behavior note (tags: nil)

Unlike detail, an explicit tags: nil is not treated as absent — it suppresses the default tag fallback (tag_object) entirely, matching the pre-migration `route.options.fetch(:tags, tag_object(route, path))` behavior. This is intentionally asymmetric with detail: `detail: nil` reads as "no separate detail text," while `tags: nil` reads as "the user explicitly opted out of default tagging." Covered by the new example in `endpoint_versioned_path_spec.rb`.

### Tests

Full suite: 515 examples, 2 failures (both pre-existing on `master`, unrelated to this change), 2 pending — no new failures. `build_body_parameter`'s unit specs are updated for the value-based signature.

### DIFF
--- a/lib/grape-swagger/doc_methods/operation_id.rb
+++ b/lib/grape-swagger/doc_methods/operation_id.rb
@@ -5,13 +5,12 @@ module GrapeSwagger
     class OperationId
       class << self
         def build(route, path = nil)
-          if route.options[:nickname]
-            route.options[:nickname]
-          else
-            verb = route.request_method.to_s.downcase
-            operation = manipulate(path) unless path.nil?
-            "#{verb}#{operation}"
-          end
+          nickname = route.nickname
+          return nickname if nickname
+
+          verb = route.request_method.to_s.downcase
+          operation = manipulate(path) unless path.nil?
+          "#{verb}#{operation}"
         end
 
         def manipulate(path)
--- a/lib/grape-swagger/endpoint.rb
+++ b/lib/grape-swagger/endpoint.rb
@@ -155,17 +155,11 @@ module Grape
     end
 
     def produces_object(route, format)
-      return ['application/octet-stream'] if file_response?(route.success) &&
-                                             !route.produces.present?
+      produces = route.produces
+      return ['application/octet-stream'] if file_response?(route.success) && !produces.present?
 
-      mime_types = GrapeSwagger::DocMethods::ProducesConsumes.call(format)
-
-      route_mime_types = %i[formats content_types produces].map do |producer|
-        possible = route.public_send(producer)
-        GrapeSwagger::DocMethods::ProducesConsumes.call(possible) if possible.present?
-      end.flatten.compact.uniq
-
-      route_mime_types.present? ? route_mime_types : mime_types
+      route_mime_types = GrapeSwagger::DocMethods::ProducesConsumes.call(produces) if produces.present?
+      route_mime_types.presence || GrapeSwagger::DocMethods::ProducesConsumes.call(format)
     end
 
     SUPPORTS_CONSUMES = %i[post put patch].freeze
--- a/lib/grape-swagger/endpoint.rb
+++ b/lib/grape-swagger/endpoint.rb
@@ -100,7 +100,7 @@ module Grape
         next if hidden?(route, options)
 
         @item, path = GrapeSwagger::DocMethods::PathString.build(route, options)
-        @entity = route.entity || route.options[:success]
+        @entity = route.entity || route.success
 
         verb, method_object = method_object(route, options, path)
 
@@ -123,7 +123,7 @@ module Grape
       method[:parameters]  = params_object(route, options, path, method[:consumes])
       method[:security]    = security_object(route)
       method[:responses]   = response_object(route, options)
-      method[:tags]        = route.options.fetch(:tags, tag_object(route, path))
+      method[:tags]        = route.options.key?(:tags) ? route.tags : tag_object(route, path)
       method[:operationId] = GrapeSwagger::DocMethods::OperationId.build(route, path)
       method[:deprecated] = deprecated_object(route)
       method.delete_if { |_, value| value.nil? }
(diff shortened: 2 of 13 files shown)


lib/grape-swagger/endpoint.rb
```diff
@@ -132,36 +132,36 @@ def method_object(route, options, path)
...
-      route.options[:security] if route.options.key?(:security)
+      route.security
     end
 
     def summary_object(route)
-      summary = route.options[:desc] if route.options.key?(:desc)
-      summary = route.description if route.description.present? && route.options.key?(:detail)
-      summary = route.options[:summary] if route.options.key?(:summary)
+      summary = route.desc if route.desc
+      summary = route.description if route.description.present? && route.detail
```
numbata, line 144: Feels like it silently changes behavior when `detail:` is explicitly set to `nil`  🤔 
Maybe add a test covering something like `desc: 'x', detail: nil` to assert the expected summary value, and document the intentional behavior change in the PR body if `nil` is meant to be treated in same way as "key absent."

lib/grape-swagger/endpoint.rb
```diff
@@ -437,7 +437,7 @@ def model_name(name)
 
     def hidden?(route, options)
       route_hidden = route.settings.try(:[], :swagger).try(:[], :hidden)
-      route_hidden = route.options[:hidden] if route.options.key?(:hidden)
+      route_hidden = route.hidden unless route.hidden.nil?
       return route_hidden unless route_hidden.is_a?(Proc)
```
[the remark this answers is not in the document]
numbata: Not a blocker. As I understand, through the `BaseRoute#method_missing`, each call delegates to `options[:hidden]`. While cheap, a double call could be unnecessary. Probably could be replaced by something like that:
```ruby
  hidden_val = route.hidden
  route_hidden = hidden_val unless hidden_val.nil?
```
But not sure about readability of it. 🤔

lib/grape-swagger/endpoint.rb
```diff
@@ -132,36 +132,36 @@ def method_object(route, options, path)
...
-      return ['application/octet-stream'] if file_response?(route.options[:success]) &&
-                                             !route.options[:produces].present?
+      return ['application/octet-stream'] if file_response?(route.success) &&
+                                             !route.produces.present?
 
       mime_types = GrapeSwagger::DocMethods::ProducesConsumes.call(format)
 
       route_mime_types = %i[formats content_types produces].map do |producer|
-        possible = route.options[producer]
+        possible = route.public_send(producer)
```
numbata: @ericproulx Do we want to have explicit public readers for them in `Grape::Route::BaseRoute`? Or we already have them?

numbata (approved):

lib/grape-swagger/endpoint.rb
ericproulx: Only `produces` is an actual Grape route option — it's in `Grape::Util::ApiDescription::DSL_METHODS`, so `Grape::Router::BaseRoute` already exposes `route.produces` as a real reader.

`formats` and `content_types` aren't Grape options. `route.formats` / `route.content_types` resolve only through `BaseRoute#method_missing`, which delegates unknown readers to `route.options[...]` — so they're really `options[:formats]` / `options[:content_types]`, grape-swagger-specific keys passed through `desc`. And they're **undocumented**: only `produces` is in the README; `formats` / `content_types` appear solely in `spec/swagger_v2/api_swagger_v2_format-content_type_spec.rb`.

So I wouldn't add Grape readers for them. I'd rather drop the undocumented `formats` / `content_types` aliases and consolidate on the documented `produces`, so we're not leaning on `method_missing`. It touches only that spec (no docs), so I'll do it in a follow-up and keep this PR a pure reader migration.

lib/grape-swagger/doc_methods/operation_id.rb
```diff
@@ -5,8 +5,8 @@ module DocMethods
     class OperationId
       class << self
         def build(route, path = nil)
-          if route.options[:nickname]
-            route.options[:nickname]
+          if route.nickname
+            route.nickname
           else
```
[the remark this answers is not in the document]
ericproulx: Good catch — cached it: `nickname = route.nickname; return nickname if nickname`.

lib/grape-swagger/endpoint.rb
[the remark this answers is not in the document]
ericproulx: Cached the reader in a local as suggested: `hidden_val = route.hidden; route_hidden = hidden_val unless hidden_val.nil?`.

lib/grape-swagger/endpoint.rb
ericproulx: Documented and tested: `detail: nil` is now treated the same as an absent `detail` — the description holds the text and there is no separate summary. Added an example in `api_swagger_v2_detail_spec.rb` and a **Behavior note** in the PR description.

lib/grape-swagger/endpoint.rb
ericproulx: Done in the follow-up #984 — `produces_object` now reads only `route.produces`, and the undocumented `formats:` / `content_types:` aliases are removed.

lib/grape-swagger/endpoint.rb
```diff
@@ -123,7 +123,7 @@ def method_object(route, options, path)
       method[:parameters]  = params_object(route, options, path, method[:consumes])
       method[:security]    = security_object(route)
       method[:responses]   = response_object(route, options)
-      method[:tags]        = route.options.fetch(:tags, tag_object(route, path))
+      method[:tags]        = route.tags || tag_object(route, path)
```
numbata: This changes explicit `tags: nil`: it previously omitted `tags`, whereas it now derives the default tag. A reader-only implementation cannot distinguish an absent `tags` key from an explicitly nil one.

Could we either preserve the old behavior with a narrow `route.options.key?(:tags)` presence check, or explicitly document this breaking change in `UPGRADE.md` (and add a regression spec for the new behavior)?

ericproulx: Confirmed — reproduced it directly with Grape 2.4.0. `route.tags` delegates to `ActiveSupport::OrderedOptions#[]`, which returns `nil` for both "key absent" and "key present with nil value", so `route.tags || tag_object(...)` can't tell an explicit `tags: nil` apart from no `tags` option at all — it always falls back to the derived tag.

Fixed with the narrow presence check you suggested:
```ruby
method[:tags] = route.options.key?(:tags) ? route.tags : tag_object(route, path)
```
This restores the old `fetch`-based behavior: explicit `tags: nil` omits the `tags` key, absent `tags` still gets the derived default. Added a regression spec ("when tags are explicitly set to nil") next to the existing custom-tags context in `spec/swagger_v2/endpoint_versioned_path_spec.rb`. Pushed in b4a72a1794e164e92ab27dada434c3c8e7b21f72.

numbata:
@ericproulx great job as usual! 👍 
Let's fix the specs and merge the PR?

ericproulx:
> @ericproulx great job as usual! 👍 Let's fix the specs and merge the PR?

on it, investigating the grape HEAD issue. Merged few things today

numbata:
I will update and merge this PR

PR #984, opened 2026-07-08
Drop the undocumented formats/content_types produces aliases

ericproulx:
Follow-up to #983 (stacked on it — until #983 merges, the diff here also shows #983's reader migration; it narrows to just this change once #983 is in).

grape-swagger let you set the swagger `produces` field three ways via `desc`: `produces:`, `formats:`, or `content_types:`. Only `produces` is documented (README `#### produces`); `formats` / `content_types` were undocumented aliases that resolved solely through `Grape::Router::BaseRoute#method_missing` delegating `route.formats` / `route.content_types` to `route.options[...]` — i.e. not real Grape route options.

This consolidates on the documented `produces`:

- `produces_object` reads only `route.produces` (a real Grape `DSL_METHODS` reader), dropping the `%i[formats content_types produces]` `public_send` loop.
- `api_swagger_v2_format-content_type_spec.rb` keeps the `produces` case (plus `use_default` and `consumes`); the `use_formats` / `use_content_types` cases are removed.
- UPGRADING note added.

This is the removal I proposed in the [#983 discussion](https://github.com/ruby-grape/grape-swagger/pull/983#discussion_r3538333406): with it, grape-swagger no longer relies on `method_missing` for route metadata.

### Tests

Full suite: 513 examples, 2 failures (both pre-existing on `master`, unrelated), 2 pending — no new failures.


numbata (approved):
Thanks for the contribution, @ericproulx!

dblock:
@numbata you have to merge it for @ericproulx he's not maintainer here

I'll hit it

PR #985, opened 2026-07-28
Fix the test suite against unreleased Grape 4.0

ericproulx:
## Summary

Two Grape 4.0 changes break the suite against `grape=HEAD`. Both are intentional upstream and documented there, so the fixes belong here.

Before this PR, `GRAPE_VERSION=HEAD bundle exec rspec` fails 23 examples; after, the suite is green on both `grape=HEAD` (4.0.0) and grape 3.3.4.

## 1. Collection custom types need `parse` (12 failures)

[ruby-grape/grape#2817](https://github.com/ruby-grape/grape/pull/2817) builds an `Array`/`Set` element coercer **eagerly**, in the coercer's constructor, so that coercers — shared across requests — create no state at request time. A collection whose element type Grape cannot coerce is therefore rejected while the `params` block is evaluated, instead of on the first request that supplies the parameter:

```
ArgumentError: type Entities::ApiError should support coercion via `[]`
```

`Entities::ApiError` is an `OpenStruct` / `Representable::Decorator` with no `parse`, so `type: Array[Entities::ApiError]` now raises as the API class loads. Fixed by giving the fixtures a one-argument `parse`, which is what the bare `type: X` fixtures already carry.

**Why this was invisible until now.** `params_array_spec` only ever `GET`s `/swagger_doc/...` — it never POSTs to the route declaring that type. Grape used to build the element coercer lazily, so nothing ever built it. Had a request supplied the parameter, grape 3.3.4 answered `400 {"error":"array_of_entities is invalid"}` — the `ArgumentError` raised inside the coercer and swallowed by the coercion validator. So this was a latent misdeclaration in the fixtures, not new breakage.

**Comment correction.** The existing fixture comments credited this to "Grape 3.2+". That's right for bare `type: X` but wrong for the collection form:

| grape | `type: X` | `Array[X]` |
| --- | --- | --- |
| 3.1.0 | OK | OK |
| 3.2.0 | `ArgumentError` | OK |
| 3.3.4 | `ArgumentError` | OK |
| 4.0 | `ArgumentError` | `ArgumentError` |

Comments now point at Grape 4.0 / #2817, and UPGRADING's custom-type bullet — which currently only covers the bare form — gains the collection case.

## 2. `StackableValues` is a read-only view (11 failures)

[ruby-grape/grape#2823](https://github.com/ruby-grape/grape/pull/2823) made `Grape::Util::StackableValues` a read-only view: `.new` now takes `(new_values, inherited_values)` and `[]=` is gone. `swagger_routing_spec` built instances directly and then wrote to them, so it fails on both counts.

`combine_namespaces` only ever *reads* keys off the object (`[](key)` returning an Array, for `.last` and `.join('/')`), so these specs now stub that narrow contract rather than constructing a real instance. That's version-agnostic and can't break again on either the constructor or the writer — the same reasoning behind the `stackable_values_double` helper already in `request_param_parsers/route_spec.rb`, which covers the other 9.

Note the fixture values changed shape slightly: the old `[]=` *pushed*, so `[:mount_path] = ['//foo/', '/bar']` actually produced `[['//foo/', '/bar']]`. The doubles use the flat form, which is what real Grape produces (`add_mount_path` pushes one string per mount level). Assertions are unchanged — `Array#join` recurses, so both shapes yield the same joined path.

**Grape's compatibility surface is unaffected:** grape-swagger's `lib` never constructs a `StackableValues`, it only reads through one (`is_a?`, `#inherited_values`, `#new_values`, `#[]`), all of which #2823 preserved deliberately.

## Verification

- Green on `grape=HEAD` (4.0.0) and grape 3.3.4.
- Example count unchanged from `master` (530, verified with `--dry-run` against both trees) — no coverage dropped.
- I could not run the grape 2.1 row locally (grape 2.1 doesn't install on Ruby 4.0.5); leaving that to CI.
- `rubocop` could not run: `.rubocop.yml` references `Style/OneClassPerFile`, which the installed RuboCop rejects. Pre-existing — it fails the same way on untouched files like `lib/grape-swagger.rb`.


ericproulx:
@numbata you need to merge this PR first to fix grape=HEAD tests.

numbata:
@dblock I will merge this and do release.

numbata (approved):

PR #986, opened 2026-08-22
Read tags without reaching into route.options

ericproulx:
`route.options.key?(:tags)` was the last read of a route's raw options Hash in `lib/`. `grep -rn "route\.options" lib` now returns nothing.

It was there to tell *"no `tags:` given"* from *"`tags: nil`"*, because Grape's `route.tags` reader answers `nil` for both. Rather than ask Grape for a way to expose that — a presence predicate, or a reader that carries it in the value — this drops the distinction, because it turns out nobody ever chose it.

## The `tags: nil` behaviour was an accident

[#523](https://github.com/ruby-grape/grape-swagger/pull/523) added route-level tags in 2016 to override path-derived grouping (a `prefix 'locations/:id'` was filing every endpoint under `locations`). It read them with:

```ruby
method[:tags] = route.options.fetch(:tags, tag_object(route))
```

`fetch(key, default)` is just the idiomatic "use the option if given" — but its default-on-absence semantics silently gave `nil` a meaning of its own: suppress the tag entirely. Nothing chose that:

- The README #523 added documents only `tags: ['tag1', 'tag2']`, and still does.
- The spec it added covers only a real list.
- **No commit in ten years of history mentions `tags: nil`.** The first is [#983](https://github.com/ruby-grape/grape-swagger/pull/983), which restored the behaviour on the assumption it was intended, after [#984](https://github.com/ruby-grape/grape-swagger/pull/984) had switched the line to `route.tags || tag_object(route, path)` a day earlier.

It is also inconsistent with every sibling option, where `nil` means *not specified*. Those only look like they check presence:

```ruby
def deprecated_object(route)
  route.options[:deprecated] if route.options.key?(:deprecated)
end
```

That guard is decorative — a single expression, so it is exactly `route.options[:deprecated]`. Same for `security_object`. `tags` was the only option whose presence changed the outcome.

And it is the less useful reading: someone writing `tags: nil` is most likely writing `tags: condition ? %w[a] : nil`, and would expect the default grouping back rather than the key silently dropped.

## The change

```diff
-method[:tags] = route.options.key?(:tags) ? route.tags : tag_object(route, path)
+method[:tags] = route.tags.presence || tag_object(route, path)
```

`tags:` is an override, and only a non-empty list overrides anything. `#presence` collapses `nil` and `[]` alike, so a blank declaration means "not specified" — exactly as it does for every other `desc` option — and falls back to the path-derived tag.

| declaration | before | after |
| --- | --- | --- |
| unset | derived tag | derived tag |
| `tags: %w[a b]` | `["a","b"]` | `["a","b"]` |
| `tags: []` | `"tags": []` | **derived tag** |
| `tags: nil` | key omitted | **derived tag** |

The `tags: []` row is a second small fix: an empty list is valid Swagger but documents nothing, and most tooling reads it as untagged anyway — so it now says what it looks like it says, "I did not set this", rather than emitting noise into the document.

## Why this shape

Grape is moving toward routes exposing readers rather than a free-form bag ([ruby-grape/grape#2857](https://github.com/ruby-grape/grape/pull/2857) removes `BaseRoute`'s `delegate_missing_to :@options`, which made every route answer every name with `nil`). Anything grape-swagger can express through `route.tags` alone is one less thing Grape has to expose to keep this gem working. Here that costs an undocumented, untested and inconsistent edge case, and buys a rule that fits the rest of `desc`.

## Backward compatibility

Neither `desc(..., tags: nil)` nor `desc(..., tags: [])` suppresses the tag any more; both derive it from the path. There is no longer a way to declare "this operation has no tags" — that was only ever reachable through the accident above. Covered in UPGRADING, with the #523 archaeology recorded so the reasoning is not lost a third time.

## Test plan

- [x] #983's example rewritten rather than deleted — `tags: nil` now asserts the derived tag — plus a sibling asserting the same for `tags: []`, so both halves of the contract are pinned.
- [x] Full RSpec suite passes locally (531 examples, 0 failures, 2 pending).
- [x] RuboCop clean (152 files).
- [x] `grep -rn "route\.options" lib` returns nothing.
- [ ] CI green.

https://claude.ai/code/session_019aqLx595ggku1E2djNDdxW

### DIFF
--- a/lib/grape-swagger/endpoint.rb
+++ b/lib/grape-swagger/endpoint.rb
@@ -123,7 +123,7 @@ module Grape
       method[:parameters]  = params_object(route, options, path, method[:consumes])
       method[:security]    = security_object(route)
       method[:responses]   = response_object(route, options)
-      method[:tags]        = route.options.key?(:tags) ? route.tags : tag_object(route, path)
+      method[:tags]        = route.tags.presence || tag_object(route, path)
       method[:operationId] = GrapeSwagger::DocMethods::OperationId.build(route, path)
       method[:deprecated] = deprecated_object(route)
       method.delete_if { |_, value| value.nil? }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 #### Fixes
 
+* [#986](https://github.com/ruby-grape/grape-swagger/pull/986): Read a route's tags through `route.tags` instead of `route.options`; a blank `tags:` (`nil` or `[]`) now means "not specified" and derives the tag from the path. See [UPGRADING](UPGRADING.md) - [@ericproulx](https://github.com/ericproulx).
 * Your contribution here.
 
 ### 2.2.0 (2026-08-06)
--- a/README.md
+++ b/README.md
@@ -648,6 +648,8 @@ end
 Tags are used for logical grouping of operations by resources or any other qualifier. To override the
 tags array, add `tags: ['tag1', 'tag2']` after the description.
 
+When `tags:` is not given, or is `nil` or empty, the tag is derived from the path.
+
 ```ruby
 namespace 'order' do
   desc 'This will be your summary', tags: ['orders']
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -1,5 +1,11 @@
 ## Upgrading Grape-swagger
 
+### Upgrading to >= 2.3.0
+
+- **A blank `desc(..., tags:)` now means "not specified".** `tags: nil` no longer suppresses the tag and `tags: []` no longer emits an empty `"tags": []`; both derive the tag from the path, as if `tags:` had been omitted. `tags:` is an override, and only a non-empty list overrides anything.
+
+  The old `nil` behaviour was never a deliberate one: [#523](https://github.com/ruby-grape/grape-swagger/pull/523) introduced route-level tags to *override* path-derived grouping and read them with `route.options.fetch(:tags, tag_object(route))`, whose default-on-absence semantics happened to give `nil` a meaning of its own. It was never documented or tested.
+
 ### Upgrading to >= 2.2.0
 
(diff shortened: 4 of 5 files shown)


lib/grape-swagger/endpoint.rb
```diff
@@ -123,7 +123,7 @@ def method_object(route, options, path)
       method[:parameters]  = params_object(route, options, path, method[:consumes])
       method[:security]    = security_object(route)
       method[:responses]   = response_object(route, options)
-      method[:tags]        = route.options.key?(:tags) ? route.tags : tag_object(route, path)
+      method[:tags]        = route.tags ? route.tags.presence : tag_object(route, path)
```
numbata: What’s the reason for using `.presence` here? If `route.tags` is absent, won’t the code go to the `else` branch anyway? 🤔

ericproulx: Sorry about that. AI slop. I think `route.tags.presence || tag_object(route, path)` is more suitable.

ericproulx:
Describing an api contains static keys and I think having dedicated readers makes more sense than always passing by `options`. IMOO, `options` is more like user defined options rather than internals.

lib/grape-swagger/endpoint.rb
numbata: @ericproulx offtop question! Did you thought about having something like:
```ruby
Grape::UNSET_VALUE = Object.new
```
To have explicit "unsetted values" and have explicit `nil` value. 🤔

ericproulx: I haven't thought of that. To be honest, I think the `nil` value is serving that purpose. I'm just curious about why `tags` is the only one calling `key?`. I haven't found a case where a user would call `tags nil` or `tags []` in a Grape::API to show an empty `tags` in swagger.

numbata (approved):

